PR #4, opened 2016-03-21
add read content method

deepu105:
This is what I was asking for in the comments


package.json
```diff
@@ -8,7 +8,7 @@
     "DSL"
   ],
   "scripts": {
-    "test": "mocha test/*"
+    "test": "mocha test/reader/*"
```
MathieuAA: It's a mistake as there will be more tests in the folder than just the reader subfolder.

module/index.js
```diff
@@ -1,10 +1,8 @@
 'use strict';
 
-var JDLParser = require('../lib/reader/jdl_reader');
+var JDLReader = require('../lib/reader/jdl_reader');
```
MathieuAA: My bad, thanks for correcting that!

package.json
deepu105: In that case we need to move the test_files location as mocha fails trying to find an index.js there

deepu105:
@MathieuAA btw are you following any specific coding style? coz I find your style bit different from normal node module style?

package.json
deepu105: may be lets make it like this
```
test
 |_specs
    |_ reader
 |_samples
```

MathieuAA:
I won't merge it as I'm quite busy, I'll discuss my problems with the PR later today, please don't merge it.

deepu105:
ok

MathieuAA:
I looked into various pseudo-coding style guides when we started JHipster UML a long time ago, but there's nothing official and the provided information is not the same every time, so I follow my own (which I find quite good =p).
As for Node modules (code, tests, etc.) they always vary a lot (especially the tests, which is weird). So, again, I follow my own. But I tend to "respect" various coding habits (wrapping apps in (function () {})(); etc..).
We can't compare JHipster's coding style with other projects as JHipster is a Yeoman generator, and I guess it must follow some strange rules... For this project, I follow the same coding habits I usually follow when working on JHipster UML.

deepu105:
Not that I find it wrong or anything just that it can be bit more simpler since its javascript and not Java/C :P

Btw there are some official style guides like the npm one itself  https://docs.npmjs.com/misc/coding-style

Since this is javascript many folks actually follow this https://github.com/airbnb/javascript (this is something im slowly trying to adapt in Jhipster, but will take some time due to the large codebase, it has nothing on yeoman to blame btw :) since its a node module it can very well follow normal js style )
This node specific one is also quite popular https://github.com/felixge/node-style-guide

MathieuAA:
Sorry for the delay, quite a lot of work to do.
Concerning the "official" style guide from NPM, my code respects it... almost. Some rules are just plain contradictory or stupid (the semi-colons is something I'd like to see less often for instance, but I doubt every platform can support that).
I like the guide proposed by AirBnb, and I really wish you luck to push it more in the codebase so that others copy this style!
However, I don't fully agree with their coding rules either... Maybe because I don't understand how V8 does some of its voodoo :-).
For instance, the "Use a leading underscore _ when naming private properties" may be the ugliest thing I've read in the guide, and I won't ever respect this rule, even if I agree 99.99% with the rest of the doc.

---

As for the PR, there's no need to have a spec subfolder in the test folder, as mocha will find every .js test file... and spec is kinda redundant with test as I use chai/expect (logical matter, not important).
You were right when you renamed the samples' folder "samples".
The test file shouldn't be named index.js for these reasons:
  - One test subfolder can have more than one test;
  - The test files should be easy to recognize;
  - As one subfolder can have more than one test file, I dont like the idea of having all my tests in one file even if you can `describe('MyClass', function...)` more than once.

deepu105:
Ideally it should take all the js files but its not working at least in
windows, but that might have something to do with the command used in npm.

I dont like having samples and tests in the same level and spec is the most
common naming style for tests in the node world, if you want sub folders
have it inside the spec folder. Did you try running the tests earlier?  Did
they work?

About coding style, when you do a personal project its fine to do whatever
style as javascript is too flexible anyway. But in a collaborative project
you have to adapt to something that all can agree. Anyway we need to put an
.editorconfig and .eslint file.

I dont have much problems with your coding style except for the use of too
many error handling classes which is kind of un necessary in javascript.

MathieuAA:
The point of having a lot of error classes is to return them to the user, or the calling code (and to help me debug without any problem)... But I have a new method for handling these, I just need to test it =).

Yes, the basic `npm test` works for my settings (see JHipster UML for example, tested in Windows and Linux distros). Does yours differ than the "normal" one is some way? I'll do some more tests tonight with different settings... Does the package.json recognize things listed as '**/*.js'?

And finally, yes. We all need to agree on some coding rules. Can you start a thread in the dev mailing list? It would be nice to know the habits of everyone so that we may decide. Thank you Deepu.

deepu105:
I have update test file names, do you have any more issues with this?

deepu105:
btw shouldnt the travis build be running `npm test`? currently it does just an `npm install`

MathieuAA:
Travis should know how to deal with a NPM package (I hope so anyway)...

MathieuAA:
Outch! You used `const`! It's a ES6 feature.

deepu105:
yes we need not support node x as Jhipster doesnt anyway

deepu105:
I meant node 0.x

MathieuAA:
See my first build =) it failed because of 2 or 3 versions that are way too old to use!

deepu105:
yes lets just support 4 onwards its waste of effort to support 0.x its too old

deepu105:
and we need to enable npm test on travis

MathieuAA:
It is enabled Deepu. Don't know what you are talking about. Check the output of your last build.
Let's not use any ES6 feature for now (even if I do want to use it myself). You should remove the `const`.

deepu105:
No Mathieu, I differ. This is a supporting lib for JHipster, we are already
using es6 in jhipster as node 4.x onwards supports it and we are
recommending node LTS so not using those to support older versions here
doesn't make sense to me.

MathieuAA:
When you say "already using", does that mean you use things as `const`, `let`, etc.? I don't see where we disagree here though. Unless you answered to my original post, I've changed my mind since then =)

deepu105:
its ok nevermind its reflected in code

Thanks & Regards,
Deepu

PR #56, opened 2016-09-01
Added mariadb database type

kingjan1999:
Currently import-jdl fails, when 'mariadb' is set as database type.


MathieuAA:
Thanks for submitting this PR!

PR #65, opened 2016-09-27
Add json to jdl conversion support

### LINKED ISSUE #64: JSON to JDL converter
Is there any way to convert existing json entities to jdl ?
If not it could be handy to have one for people moving their project to jdl

cbornet:
One of the test does a json -> jdl -> json and compares the initial and produced json. This is useful to detect issues during one of the stages. And actually I found bugs which are fixed in this PR:
 * the JDL field `comment` was not converted to json `javadoc`
 * the JDL `noFluentMethod` was output in the json together with `"fluentMethods": false`
 * the JDL `angularSuffix` was not converted to json `angularJSSuffix`
 * the json `searchEngine` was not converted to JDL `search`

Fix #64

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,8 @@ node_modules
 test/temp
 npm-debug.log
 out/
-.jhipster/
+/.jhipster/
+jhipster-jdl.jh
 .sonar
 coverage/*
 sonar-project.properties
--- a/lib/parser/json_parser.js
+++ b/lib/parser/json_parser.js
@@ -1,0 +1,194 @@
+'use strict';
+
+const _ = require('lodash'),
+    JDLObject = require('../core/jdl_object'),
+    JDLEntity = require('../core/jdl_entity'),
+    JDLField = require('../core/jdl_field'),
+    JDLValidation = require('../core/jdl_validation'),
+    JDLEnum = require('../core/jdl_enum'),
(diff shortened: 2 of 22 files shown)


.gitignore
```diff
@@ -3,7 +3,8 @@ node_modules
 test/temp
 npm-debug.log
 out/
-.jhipster/
+/.jhipster/
+jhipster-jdl.jh
```
cbornet: it is generated during tests and normally deleted

lib/parser/json_parser.js
```diff
@@ -0,0 +1,184 @@
...
+ * jdl: JDLObject to which the parsed entities are added. If undefined a new JDLObject is created.
+ * returns the JDLObject
+ */
+function parseEntities(entities, jdl) {
+  if (!entities) {
+    throw new buildException(
+      exceptions.NullPointer,
+      'Entities have to be passed to be converted.');
+  }
+  if (jdl === undefined) {
```
MathieuAA: What if I pass `null`?

cbornet: If you pass null, you're voluntarily asking for trouble and you'll get a NPE quite quickly! Throwing a buildException would be cleaner I guess.

cbornet: I've added the null check + exception

PR #77, opened 2016-10-13
Parse server options from jhipster config field instead of full .yo-rc.json

cbornet:
This is more consistent.
In generator-jhipster, this is `this.config` and you don't need to wrap it.


MathieuAA:
Well, if you've tested it and seen it work, then Thanks =)

Issue #81, opened 2016-10-26
Comments in relationships not working


cbornet:
* JDL to JSON : looks like a bug since there is code to support it
* JSON to JDL : not implemented yet

```
yo jhipster:import-jdl /home/vagrant/workspace/jhipster-core/test/test_files/complex_jdl.jdl
```
Expected to see `"javadoc": "A relationship"` in Department.json but it's not there

MathieuAA:
I didn't know it was the international JHipster issue day!!

cbornet:
I'm reopening because the JDL to JSON conversion still needs to be fixed

MathieuAA:
Oh! Yes of course!

PR #82, opened 2016-10-26
Support javadoc in relationships for Json to JDL conversion

cbornet:
See #81

### DIFF
--- a/lib/parser/json_parser.js
+++ b/lib/parser/json_parser.js
@@ -119,6 +119,7 @@ function addRelationshipsToJDL(jdl, entities) {
       }
       let injectedFieldInTo;
       let isInjectedFieldInToRequired;
+      let commentInTo;
       let otherEntity = entities[_.upperFirst(relationship.otherEntityName)];
       if (otherEntity === undefined) {
         continue;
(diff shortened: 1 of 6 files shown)


lib/parser/json_parser.js
```diff
@@ -152,7 +155,9 @@ function addRelationshipsToJDL(jdl, entities) {
             injectedFieldInFrom: injectedFieldInTo,
             injectedFieldInTo: injectedFieldInFrom,
             isInjectedFieldInFromRequired: isInjectedFieldInToRequired,
-            isInjectedFieldInToRequired: relationship.relationshipValidateRules
+            isInjectedFieldInToRequired: relationship.relationshipValidateRules,
+            commentInFrom: commentInTo,
```
sendilkumarn: is this intentional ?

cbornet:
I've also fixed the toString() methods for multi-line comments.
Question: should we favor `/* some comment */` comments when there is only one line in comment ?
I personally prefer it since it's more compact

MathieuAA:
Frankly, users could do both. But if we generate to JDL, the point is to propose the best version... and I agree, this version is better _in this case_ as the comment is just a one-liner here.

PR #88, opened 2016-11-28
drive multiline javadoc unchanged from JDL Studio to generated class

kaidohallik:
this change is tested with PR jhipster/generator-jhipster#4561
advisable is merge this PR after jhipster/generator-jhipster#4561 is merged

fix jhipster/jhipster-core#87


MathieuAA:
Looks good to me, thanks!

PR #90, opened 2016-11-29
entity field inline javadoc support

kaidohallik:
if both javadoc is entered:
- javadoc above the field
- inline javadoc
then these 2 comments are merged

fix jhipster/jhipster-core#89


MathieuAA:
There is no issue here.

PR #93, opened 2016-12-02
Add mssql database type and reserved words

ruddell:
Tested with the JDL Studio sample and tested to make sure MSSQL reserved keywords fail.

Related to jhipster/generator-jhipster#4518 and jhipster/generator-jhipster#4589


MathieuAA:
Thanks Jon!

PR #102, opened 2017-02-03
Add text and type to reserved words for Angular

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/5079
Fix https://github.com/jhipster/generator-jhipster/issues/5081


jdubois:
Yes, but I don't understand why... For type I can imagine, but text???

pascalgrimaud:
I don't understand too :laughing: :rofl:

jdubois:
OK, then:

- First, I would like to understand what happens here, as there are probably more keywords like those
- Second, I would like to know if we could find some way to still be able to use those names: we already have too many reserved keywords... BTW we could probably use the SQL reserved keywords, using backquotes, I need to check those

gmarziou:
My PR is based on list of typescript reserwed words found [here (first list)](https://narayanatechnicalworld.blogspot.fr/2016/05/typescript-reserved-words.html) for which I tested few of them and they can't be used as class properties ether defined in constructor or in class body.

text and type are a different issue, tsc will not, report an error if you use them as a property, I tested it, code below compiles fine.

~~~typescript
class Greeter {
    any;
    string;
    class;
    text;
    type;

    constructor(
        public greeting: string,
        public boolean: string) { }
};
~~~~

gmarziou:
The only way to avoid this would be to use a convention for naming properties maybe with a `_` prefix but IMO this would be ugly code.

I'd prefer to exclude these words so that user knows they have to find a synonym or a different way to write it like in java we often use `clazz` or `klass` to avoid `class`. So maybe we could provide a list of synonyms for our 2 supported compiled languages: java and typescript.

The text and type issue should be investigated further.

gmarziou:
Typescript has some mechanism to escape these words for property names but it does not  work for variable names so we may encounter other issues down the road.

https://stackoverflow.com/questions/38346223

Many reserved words are common to both java and typescript, so I guess we won't forbid a lot of new words.

The largest list comes from Oracle so we should probably use backquotes as you suggested.

gmarziou:
For text, the problem comes from `Response` which extends `Body` which has text() method.

Our generated code is wrong, I will PR it.

pascalgrimaud:
Oh good catch @gmarziou ! I close this PR.
I will made another one for the tests I added

PR #104, opened 2017-02-03
Add Typescript reserved words that cannot be used as an Identifier

gmarziou:
See https://narayanatechnicalworld.blogspot.fr/2016/05/typescript-reserved-words.html


gmarziou:
Similar to issue jhipster/generator-jhipster#5081

jdubois:
@gmarziou have a look at #102 you are trying to do the same thing

And I understand your solution much better, but it doesn't have the "text" or "type" keywords, so there is still a big question here

gmarziou:
My PR is based on list of typescript reserwed words found [here (first list)](https://narayanatechnicalworld.blogspot.fr/2016/05/typescript-reserved-words.html) for which I tested few of them and they can't be used as class properties ether defined in constructor or in class body.

text and type are a different issue, tsc will not, report an error if you use them as a property, I tested it.

~~~
class Greeter {
    any;
    string;
    class;

    constructor(
        public greeting: string,
        public boolean: string) { }
};
~~~~

gmarziou:
The only way to avoid this would be to use a convention for naming properties maybe with a `_` prefix but IMO this would be ugly code.

I'd prefer to exclude these words so that user knows they have to find a synonym or a different way to write it like in java we often use `clazz` or `klass` to avoid `class`.

PR #117, opened 2017-04-23
Pagination is now forbidden when having a cassandra app

MathieuAA:
@pascalgrimaud Do you agree with the changes I propose? Especially the fact that I'd rather exit from the JDL processing process than warn the user (and not exit).


pascalgrimaud:
Yes agree to exit!
And LGTM (just read your code, I didnt test it)

PR #134, opened 2017-07-13
fix example JDL DTO option

ruddell:
The DTO option needs to be enabled for any entities with a relationship to an entity using DTOs.  Employee has a relationship with every entity, so I replaced it with `*`

Related to https://github.com/jhipster/generator-jhipster/issues/6096


MathieuAA:
Thanks guys

PR #135, opened 2017-07-14
Fix validations rules

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/6010

For example, when `name=min` and `value=0`, the result is `min` instead of `min(0)`
It's because the condition `if (this.value)` will evaluate to `true` if value is not:
- null
- undefined
- NaN
- empty string
- 0
- false

So the 0 need to be excluded here

Hope I won't break something, I'm trusting on your tests @MathieuAA 
But plz, check closely if there is any impact

### DIFF
--- a/lib/core/jdl_validation.js
+++ b/lib/core/jdl_validation.js
@@ -37,7 +37,7 @@ class JDLValidation {
 
   toString() {
     let string = `${this.name}`;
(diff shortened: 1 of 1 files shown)


MathieuAA:
I think it's OK Pascal, thanks.

Issue #141, opened 2017-08-28
Enriching the JDL grammar


MathieuAA:
From the generator's original post [here](https://github.com/jhipster/generator-jhipster/issues/6275)

colameo:
Please keep in mind, that changes in the JDL grammar affects the community which uses jhipster-ide. Since the ide features were introduced we have [300-500 installation](https://marketplace.eclipse.org/content/jhipster-ide/metrics#group-metrics-tab) per month and as you can imagine a change in the JDL would affect those people...

Beside of that, I also like the idea to improve the JDL grammar ;-) and thought to implement a kind of "Xjdl" (which stands for eXtended JDL) with feature similar to what you discussed in the issue [#6275](https://github.com/jhipster/generator-jhipster/issues/6275).

So I'm looking forward to this...

deepu105:
Colameo, the changes will be mostly backward compatible I believe

MathieuAA:
Yes. The only breaking change I've ever made was about regexes (the separator is now `/` instead of `'`).

colameo:
@MathieuAA - ok that's good, that means we do not have pressure to integrate all changes in sync with yours.

agoncal:
@deepu105 I liked the `abstract` and the `type microservice`, this leaves other possibilities.

Deepu, I didn't get the difference between `services` (plural) and `service` singular in your post : https://github.com/jhipster/generator-jhipster/issues/6275#issuecomment-325188239. You wrote `entity MyOnlyEntity` (singular)  and `services MyOnlyService` (plural). Is this a typo ?

```
 entity MyOnlyEntity extends C {     // one entity of the service extends one from commonConfig.jdl
    // this will have the field name generated as it extends entity C
    String code
  }

  services MyOnlyService {
    Customer findByName(String name)
  }
```

MathieuAA:
@agoncal it must be.
@jhipster/developers, @bd82, @hakandilek My main goal this week will be to get the app generation up and running. This means being able to generate apps from one or more JDL files, **without generating entities right after (this comes after)**. 

@deepu105 Do you think it will be easy to update the subgen to make it generate an application?

deepu105:
[...]

PR #142, opened 2017-08-28
WIP - POC with Chevrotain Parser.

bd82:
** DO NOT MERGE WIP**

### DIFF
--- a/lib/dsl/chev_grammar.js
+++ b/lib/dsl/chev_grammar.js
@@ -1,0 +1,133 @@
+const chevrotain = require('chevrotain');
+const _ = require('lodash');
+
+// ----------------- lexer -----------------
+const Lexer = chevrotain.Lexer;
+
+
+const tokens = {};
+
+function createToken(config) {
+  tokens[config.name] = chevrotain.createToken(config);
+}
+
+createToken({
+  name: 'WhiteSpace',
+  // TODO: uncertain why the original grammar disallowed newlines
+  // TODO: in rules: "fieldDeclList: "and validationList"
+  pattern: /\s+/,
+  group: Lexer.SKIPPED,
+  line_breaks: true
+});
+
+// Comments
+createToken({
+  name: 'Comment',
+  pattern: /\/\*[^]*?\*\//,
+  // a completely ignored comment, will not be a Javadoc comment
+  group: 'comments',
+  line_breaks: true
+});
+
+// Constants
+// Application constants
+createToken({ name: 'APPLICATION', pattern: 'application' });
+createToken({ name: 'BASE_NAME', pattern: 'baseName' });
+createToken({ name: 'PATH', pattern: 'path' });
+createToken({ name: 'PACKAGE_NAME', pattern: 'packageName' });
+createToken({ name: 'AUTHENTICATION_TYPE', pattern: 'authenticationType' });
+createToken({ name: 'HIBERNATE_CACHE', pattern: 'hibernateCache' });
+createToken({ name: 'CLUSTERED_HTTP_SESSION', pattern: 'clusteredHttpSession' });
+createToken({ name: 'WEBSOCKET', pattern: 'websocket' });
+createToken({ name: 'DATABASE_TYPE', pattern: 'databaseType' });
+createToken({ name: 'DEV_DATABASE_TYPE', pattern: 'devDatabaseType' });
+createToken({ name: 'PROD_DATABASE_TYPE', pattern: 'prodDatabaseType' });
+createToken({ name: 'USE_COMPASS', pattern: 'useCompass' });
+createToken({ name: 'BUILD_TOOL', pattern: 'buildTool' });
+createToken({ name: 'SEARCH_ENGINE', pattern: 'searchEngine' });
+createToken({ name: 'ENABLE_TRANSLATION', pattern: 'enableTranslation' });
+createToken({ name: 'APPLICATION_TYPE', pattern: 'applicationType' });
+createToken({ name: 'TEST_FRAMEWORK', pattern: 'testFrameworks' });
+createToken({ name: 'LANGUAGES', pattern: 'languages' });
+createToken({ name: 'SERVER_PORT', pattern: 'serverPort' });
+createToken({ name: 'ENABLE_SOCIAL_SIGN_IN', pattern: 'enableSocialSignIn' });
+createToken({ name: 'USE_SASS', pattern: 'useSass' });
+createToken({ name: 'JHI_PREFIX', pattern: 'jhiPrefix' });
+createToken({ name: 'MESSAGE_BROKER', pattern: 'messageBroker' });
+createToken({ name: 'SERVICE_DISCOVERY_TYPE', pattern: 'serviceDiscoveryType' });
+createToken({ name: 'CLIENT_PACKAGE_MANAGER', pattern: 'clientPackageManager' });
+createToken({ name: 'CLIENT_FRAMEWORK', pattern: 'clientFramework' });
+createToken({ name: 'NATIVE_LANGUAGE', pattern: 'nativeLanguage' });
+createToken({ name: 'FRONT_END_BUILDER', pattern: 'frontendBuilder' });
+createToken({ name: 'SKIP_USER_MANAGEMENT', pattern: 'skipUserManagement' });
+// skipClient & skipServer are already defined
+createToken({ name: 'TRUE', pattern: 'true' });
+createToken({ name: 'FALSE', pattern: 'false' });
+// Entity constants
+createToken({ name: 'ENTITY', pattern: 'entity' });
+createToken({ name: 'RELATIONSHIP', pattern: 'relationship' });
+createToken({ name: 'ENUM', pattern: 'enum' });
+// Relationship types
+createToken({ name: 'ONE_TO_ONE', pattern: 'OneToOne' });
+createToken({ name: 'ONE_TO_MANY', pattern: 'OneToMany' });
+createToken({ name: 'MANY_TO_ONE', pattern: 'ManyToOne' });
+createToken({ name: 'MANY_TO_MANY', pattern: 'ManyToMany' });
+
+// Options
+createToken({ name: 'ALL', pattern: 'all' });
+createToken({ name: 'STAR', pattern: '*' });
+createToken({ name: 'FOR', pattern: 'for' });
+createToken({ name: 'WITH', pattern: 'with' });
+createToken({ name: 'EXCEPT', pattern: 'except' });
+createToken({ name: 'NO_FLUENT_METHOD', pattern: 'noFluentMethod' });
+createToken({ name: 'DTO', pattern: 'dto' });
+createToken({ name: 'PAGINATE', pattern: 'paginate' });
+createToken({ name: 'SERVICE', pattern: 'service' });
+createToken({ name: 'MICROSERVICE', pattern: 'microservice' });
+createToken({ name: 'SEARCH', pattern: 'search' });
+createToken({ name: 'SKIP_CLIENT', pattern: 'skipClient' });
+createToken({ name: 'SKIP_SERVER', pattern: 'skipServer' });
+createToken({ name: 'ANGULAR_SUFFIX', pattern: 'angularSuffix' });
+
+// validations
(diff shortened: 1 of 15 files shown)


bd82:
Hi. I'm playing around with re-implementing the JDL parser using the [Chevrotain](https://github.com/SAP/chevrotain) parsing library.

This is related to the technology choice discussion in #141 and the 
the preceding discussion in https://github.com/jhipster/generator-jhipster/issues/6275
This may need its own separate issue...

This PR currently contains a lexer for JDL implemented using Chevrotain RegExp based lexer engine.

There are some minor changes from the original JDL pegjs implementation documented in the comments.

The next step is to try and convert the grammar itself (just syntax, no AST building).

bd82:
Thanks @deepu105 .

Being an internal DSL the grammar itself is imho a little uglier than pure [EBNF](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form) style syntax, but still **highly readable** by using vertical spacing.

Example:
```javascript
    // comments will be handled outside(after) the parser in this implementation.
    $.RULE('entityBody', () => {
      $.CONSUME(t.LCURLY);
      $.AT_LEAST_ONE(() => {
        $.SUBRULE($.fieldDec);
      });
      $.CONSUME(t.RCURLY);
    });

    $.RULE('fieldDec', () => {
      $.CONSUME(t.NAME);
      $.SUBRULE($.type);
      // Short form for: "(X(,X)*)?"
      $.MANY_SEP({
        SEP: t.COMMA,
        DEF: () => {
          $.SUBRULE($.validation);
        }
      });
      $.CONSUME(t.RCURLY);
    });
```

With pegjs it would look something like: (taken from existing Grammar)
Which is a-lot more "horizontal"...
```
entityBody
  = '{' SPACE* fdl:fieldDeclList SPACE* '}' { return fdl; }
  / '' { return []; }

fieldDeclList
  = SPACE* com:comment? SPACE* f:FIELD_NAME SPACE_WITHOUT_NEWLINE* t:type SPACE_WITHOUT_NEWLINE* vl:validationList? SPACE_WITHOUT_NEWLINE* com2:comment? SPACE_WITHOUT_NEWLINE* ','? SPACE* fdl:fieldDeclList {
    return addUniqueElements([{ name: f, type: t, validations: vl, javadoc: com || com2 }], fdl );
  }
  / '' { return []; }
```

Which is in theory could be prettier, but because many things were added:
1. "**SPACE***" everywhere because pegjs cannot ignore tokens.
2. labels (**vl:**validationlist?)
3. JS code snippets to execute (semantic actions).
   * With Chevrotain you can **optionally** implement the semantic actions **outside** the grammar.

The end result is **much less readable** imho as there is **no separation of concerns** here..

But it is not just about readability, it is also maintainability.
With chevrotain you can place a breakpoint anywhere in your grammar and just debug it 
**as any other javaScript code you write**.  :)

bd82:
Please have a look at a subset of the grammar implemented in latest commit.

The next step would be to add a few tests to demonstrate capabilities on this grammar subset.
* Autocomplete support.
* Building an AST using external semantic actions (not embedded in the grammar as with pegjs).
* Linking jsdocs comments back to the AST.
* Multiple Syntax Errors (for a single input text).
* Error Recovery.
* Extracting data required to implement a JDL code formatter.
* Syntax Diagrams.

**Hopefully** I will have time to implement some of these tomorrow.

lib/dsl/chev_grammar.js
```diff
@@ -91,12 +102,18 @@ createToken({ name: 'ANGULAR_SUFFIX', pattern: 'angularSuffix' });
...
-createToken({ name: 'MINLENGTH', pattern: 'minlength' });
-createToken({ name: 'MAXLENGTH', pattern: 'maxlength' });
-createToken({ name: 'MINBYTES', pattern: 'minbytes' });
-createToken({ name: 'MAXBYTES', pattern: 'maxbytes' });
-createToken({ name: 'MAX', pattern: 'max' });
-createToken({ name: 'MIN', pattern: 'min' });
+
+// HIGHLIGHT:
+// "MIN_MAX_KEYWORD" is an "abstract" token which other concrete tokens inherit from.
+// This can be used to reduce verbosity in the parser.
```
MathieuAA: How so?

lib/dsl/chev_grammar.js
```diff
@@ -114,20 +131,135 @@ createToken({ name: 'INTEGER', pattern: /-?\d+/ });
...
+      // if we would be use some JS combinator.
+      // debugger; 
+      $.CONSUME(t.REGEX);
+      $.CONSUME(t.RPAREN);
+    });
+
+    // very important to call this after all the rules have been defined.
+    // otherwise the parser may not work correctly as it will lack information
+    // derived during the self analysis phase.
+    Parser.performSelfAnalysis(this);
```
MathieuAA: This is during the object's construction-time. All this. Why not having another way?
One file is enough, but putting everything in the constructor isn't really something I look forward to maintaining, even if the improvement of using using Chevrotain over PegJS is obvious. Why not, for instance, use a factory of some sort (a function that calls other functions to build the parser instance)?

lib/dsl/chev_grammar.js
bd82: It is used in the validation rule instead of specifying the six different keywords.
https://github.com/jhipster/jhipster-core/pull/142/files#diff-802ee05eaf770a8bbbc2fe7ef13a3efaR233

bd82: Here is the corresponding section in the existing grammar:
https://github.com/jhipster/jhipster-core/blob/fd8f71200abcb4ffccb7063a4312641289c090c4/lib/dsl/grammar.txt#L511-L522

The token inheritance does not have to be used.
It is an example for what is possible and could be considered...

lib/dsl/chev_grammar.js
bd82: > Why not having another way?

### Answer
The syntax I prefer relies on using **class fields** ESNext syntax.
https://github.com/tc39/proposal-class-fields
But this is not yet supported afaik (currently stage 3 proposal).
I suppose Babel will support this at some point:
https://github.com/babel/proposals/issues/12

TypeScript has something similar which already works **now**.
See [this example](https://github.com/SAP/chevrotain/blob/master/examples/implementation_languages/typescript/typescript_json.ts#L69):
This is similar the "official" API I'm aiming for, but may need to wait for ES2018 for that. :(


### Alternative
Anyhow as it is all just plain JavaScript you can define it (mostly) however you want...
An extreme example would be this completely different DSL for specifying Chevrotain grammars
https://github.com/kristianmandrup/chevrotain-rule-dsl

I'm am a bit too tired to think a concrete alternative syntax right now.
But I believe one should be possible even with ES6, perhaps you have a suggestion?
The constraints are:
1. Parser.**performSelfAnalysis** must be called after the rules have been defined.
    * As it relies on side effects of creating the rules.

2. The **RULE** calls must be called in the context of  the parser instance (this).

And if it helps normally you only use a single parser instance and **reset** it's internal state before each use.

### Future / Long term.
There is also an [open issue](https://github.com/SAP/chevrotain/issues/480) for better support of custom APIs for building Chevrotain parsers.
And I'm hoping in the long term to support **three** different API styles (same as Mocha/Chai have different APIs using the same underlying engine). 
1. Low Level Hand-Built style.
2. Combinator Style, fluent DSL.
3. EBNF generator style (Like pegjs).

bd82: Here is a really quick and dirty factory style hack.
https://github.com/SAP/chevrotain/blob/5235a12da1818aaf2ac075cd4326d46e46da15fc/examples/grammars/json/json.js#L95-L126

And here are the rules defined **outside** the constructor.
https://github.com/SAP/chevrotain/blob/5235a12da1818aaf2ac075cd4326d46e46da15fc/examples/grammars/json/json.js#L129-L180

I don't think this should be part of Chevrotain's official API
as I would rather wait for class fields proposal, but it can be cleaned up and reused 
by end users if needed...

Hope this example demonstrates how due to Chevrotain being a **library**
instead of a code generator makes it much more malleable to customization. 😄

deepu105:
Wow great work @bd82 and thanks

bd82:
Happy to help @deepu105 😄 

Latest commit cleaned up a bit and has a small **parser** test (happy path) which you can debug.
I plan to add more scenarios (as specs) tomorrow.

MathieuAA:
Wow. Nice!

bd82:
Added some more examples for both **syntactic content assist**
and for **error recovery / fault tolerance.**

Additionally syntax diagrams can be generated from the grammar.
This can be useful as part of a documentation site.
[Diagrams of the current sub-grammar](https://rawgit.com/bd82/jhipster-core/b2d50a745ad8777c5a9ea51ac053f816981aa9bd/lib/dsl/poc/diagrams.html)

bd82:
I think there is now enough content and E2E flow in this POC to be worth **discussing and reviewing**.
I will create a separate Issue for this (tomorrow) with specific highlights and links to the source code
to help review this fairly large number of lines.

bd82:
Added the discussion issue:
https://github.com/jhipster/jhipster-core/issues/144

lib/dsl/poc/lexer.js
```diff
@@ -0,0 +1,168 @@
...
+// a single token type. This is because they would cause ambiguities
+// when the lexing stage is separated from the parsing stage.
+// They restrictions on the names should be implemented as semantic checks
+// That approach could also provide a better experience in an Editor
+// As semantic checks don't require fault tolerance and recovery like
+// syntax errors do.
+const NAME = chevrotain.createToken({ name: 'NAME', pattern: namePattern });
+
+
+function createToken(config) {
```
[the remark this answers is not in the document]
bd82: This is subjective style question, Choose whichever you prefer...
There are actually [four/five possible argument types](http://sap.github.io/chevrotain/documentation/0_32_1/interfaces/_chevrotain_d_.itokenconfig.html#pattern) documented in the Docs

In addition using strings is just a convenience style, if you prefer conformity
and no runtime type checks you can replace the strings with regExps, eg:
```javascript
// both of these are equivalent.
createToken({ name: 'ENTITY', pattern: "entity" });
createToken({ name: 'ENTITY', pattern: /entity/ });
```

lib/dsl/poc/lexer.js
```diff
@@ -0,0 +1,168 @@
...
+createToken({ name: 'RCURLY', pattern: '}' });
+createToken({ name: 'LSQUARE', pattern: '[' });
+createToken({ name: 'RSQUARE', pattern: ']' });
+createToken({ name: 'COMMA', pattern: ',' });
+createToken({ name: 'COLON', pattern: ':' });
+createToken({ name: 'EQUALS', pattern: '=' });
+createToken({ name: 'DOT', pattern: '.' });
+
+// Imperative the "NAME" token will be added after all the keywords to resolve keywords vs identifier conflict.
+tokens.NAME = NAME;
```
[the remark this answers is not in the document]
bd82: There are many rules as with any complex domains.
Most of these rules are not specific to Chevrotain but to general writing of parsers.
E.g: [Keywords vs Identifiers in Antlr4](https://github.com/antlr/grammars-v4/blob/master/java/Java.g4#L972-L976)

In general Chevrotain tries to detect these issues and provide useful error messages
or even links with detailed instructions[[1]](https://github.com/SAP/chevrotain/blob/master/docs/resolving_grammar_errors.md#-common-prefix-ambiguities) [[2]](https://github.com/SAP/chevrotain/blob/master/docs/resolving_lexer_errors.md) on how to resolve those.

But not everything can (currently) be automatically detected (such as keyword vs identifiers).But you just gave me [an idea](https://github.com/SAP/chevrotain/issues/557) how to automatically detect keywords vs Identifiers!
👍

deepu105:
Awesome work. Though I have no idea how the formatting could be utilized with codeMirror we use in JDL studio, personally formatting is not a requirement at this point but if easy to integrate it would be cool as well

bd82:
>  have no idea how the formatting could be utilized with codeMirror we use in JDL studio.

Neither do I 😄. 
The important point is that this approach **enables future Editor tooling** extensions
by keeping all the syntactic data, and more importantly it enables those **without** needing to modify the parser.

deepu105:
That escalated quickly :smile:

PR #152, opened 2017-10-08
Travis node versions

bd82:
Only test on "official" versions of node.

- 4 is in LTS until April 2018
- 7 is not a supported version.
- Test on latest version 6, not multiple versions of 6.

See: https://github.com/nodejs/Release


.travis.yml
```diff
@@ -1,7 +1,5 @@
 language: node_js
 node_js:
   - "8"
-  - "7"
   - "6"
-  - "6.10"
-  - "6.11"
+  - "4"
```
deepu105: we dont support node 4 anymore so please remove it

deepu105: lets test only on 6 and 8

bd82:
Removed 4

It will no longer be supported in a few months anyhow ;)

PR #153, opened 2017-10-08
Ast visit missing

bd82:


bd82:
hmm closing, this should not contain 3 commits wierd

PR #154, opened 2017-10-08
Added missing ast visitor methods.

bd82:
- empty implementation.

### DIFF
--- a/lib/dsl/poc/ast_builder.js
+++ b/lib/dsl/poc/ast_builder.js
@@ -19,12 +19,10 @@ const BaseJDLCSTVisitor = new JDLParser().getBaseCstVisitorConstructor();
  * 1. ctx.NAME[0].image --> ctx.NAME[0] ? ctx.NAME[0].image : "???"
  */
 class JDLAstBuilderVisitor extends BaseJDLCSTVisitor {
-  // constructor() {
-  //   super();
-  //   // TODO: commented out because the validation would fail as we are missing methods
-  //   // due to new grammar rules.
-  //   // this.validateVisitor();
-  // }
+  constructor() {
+    super();
+    this.validateVisitor();
+  }
 
   prog(ctx) {
     return {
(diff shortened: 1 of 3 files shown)


lib/dsl/poc/ast_builder.js
```diff
@@ -102,6 +100,90 @@ class JDLAstBuilderVisitor extends BaseJDLCSTVisitor {
...
+
+  dtoDecl(ctx) {
+    // TODO: implement
+  }
+
+  entityList(ctx) {
+    // TODO: implement
+  }
+
+  method(ctx) {
```
MathieuAA: Can it be more precise (the name)?

lib/dsl/poc/ast_builder.js
```diff
@@ -102,6 +100,90 @@ class JDLAstBuilderVisitor extends BaseJDLCSTVisitor {
...
+
+  noFluentMethod(ctx) {
+    // TODO: implement
+  }
+
+  filterDecl(ctx) {
+    // TODO: implement
+  }
+
+  negativeForDecl(ctx) {
```
MathieuAA: What is it?

lib/dsl/poc/ast_builder.js
bd82: I will try to find a better name

lib/dsl/poc/ast_builder.js
bd82: This is the one I need to rename to filterDef
https://github.com/jhipster/jhipster-core/blob/master/lib/dsl/poc/parser.js#L293

lib/dsl/poc/ast_builder.js
bd82: Method is the name used in the previous grammar (but locally scoped).
https://github.com/jhipster/jhipster-core/blob/master/lib/dsl/grammar.txt#L491-L495

Do you have a suggestion for a better name?
I can also remove this grammar rule (method) completely.
But this would mean a tiny bit more complexity in the CST building.
lets deal with it in a separate PR.

bd82:
I renamed to filterDef. I have an idea how to improve the "method" name and will do so later with a different commit.

bd82:
> The indentation shouldn't be of 4 (or is it 8?), it should only be of 2.

That bad indentation file is just some **generated** code for drawing the diagrams, normally
it should not be merged into master. I am only adding it to allow sharing the diagrams easily
via gitraw **during development.**

I can remove it completely now or leave it during development, whichever you choose.

MathieuAA:
Oh, ok then! Let's keep it

bd82:
Great. let me know if anything else needs to be fixed or we can merge because I am starting to get merge conflicts with myself :)

PR #155, opened 2017-10-08
Example of multiMode lexer for "freestyle" line

bd82:
following "path" keyword in JDL application config.

### DIFF
--- a/test/spec/grammar/lexer_test.js
+++ b/test/spec/grammar/lexer_test.js
@@ -51,4 +51,36 @@ describe('Chevrotain Lexer POC', () => {
     expect(lexResult.tokens).to.have.lengthOf(9,
       'All 9 tokens should have been lexed even thought "@@@" caused a syntax error');
   });
+
+
+  it('Can lex a simple JDL with freeStyle "path line"', () => {
+    // invalid token but the lexing should continue
+    const input = `
+      application {
+        config {
+          path ../../toto!!!
+          languages en, fr
+        }
+    }`;
+    const lexResult = JDLLexer.tokenize(input);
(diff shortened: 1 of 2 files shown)


test/spec/grammar/lexer_test.js
```diff
@@ -51,4 +51,36 @@ describe('Chevrotain Lexer POC', () => {
...
       'All 9 tokens should have been lexed even thought "@@@" caused a syntax error');
   });
+
+
+  it('Can lex a simple JDL with freeStyle "path line"', () => {
+    // invalid token but the lexing should continue
+    const input = `
+      application {
+        config {
+          path ../../toto!!!
```
deepu105: would this also work with path '../../toto'

bd82: yes it would "eat" everything after the path in the **same** line.

I don't actually recommend this as mentioned in https://github.com/jhipster/jhipster-core/issues/149#issuecomment-334996508.

But this allows keeping backward compatibility with the old implementation.

MathieuAA: I didn't have time to answer yesterday, but if adding some kind of separator is needed, then let's.
I agree we've add some "parsing-code smells" in our former parsing system, but if we can get rid of them we should!

deepu105: Yes, and I don't think this path syntax was used in existing JDL anyway so this is new and we can have separator or quotes if we want

bd82: Sounds good,

As long as we don't mind a small breaking change to clean up the grammar
in this specific case 😄.

PR #157, opened 2017-10-10
Support Keywords as Identifiers

bd82:
at the syntactic analysis level (pure parsing).

To the best of my understanding this is compatiable
with how the pegjs parser worked.

This does not mean that "entity enum {...}" is valid.
Only that it will pass "pure" parsing and should fail
in the reserved words check. (reserved_words.js).

### DIFF
--- a/lib/dsl/poc/lexer.js
+++ b/lib/dsl/poc/lexer.js
@@ -33,7 +33,7 @@ function createToken(config) {
   // a Concise way to resolve the problem without manually adding the "longer_alt" property dozens of times.
   if (_.isString(config.pattern) && namePattern.test(config.pattern)) {
     config.longer_alt = NAME;
-    // 'application' IS-A KEYWORD which in turn IS-A NAME --> all keywords are now valid identifiers.
+    // 'application' IS-A KEYWORD which in turn IS-A NAME
     config.parent = KEYWORD;
   }
 
--- a/lib/dsl/poc/lexer.js
+++ b/lib/dsl/poc/lexer.js
@@ -17,6 +17,15 @@ const namePattern = /[a-zA-Z_][a-zA-Z_\d]*/;
 // syntax errors do.
 const NAME = chevrotain.createToken({ name: 'NAME', pattern: namePattern });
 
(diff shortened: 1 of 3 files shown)


lib/dsl/poc/lexer.js
```diff
@@ -17,13 +17,24 @@ const namePattern = /[a-zA-Z_][a-zA-Z_\d]*/;
...
+
 function createToken(config) {
   // JDL has a great many keywords. Keywords can conflict with identifiers in a parsing
   // library with a separate lexing phase.
   // See: https://github.com/SAP/chevrotain/blob/master/examples/lexer/keywords_vs_identifiers/keywords_vs_identifiers.js
   // a Concise way to resolve the problem without manually adding the "longer_alt" property dozens of times.
   if (_.isString(config.pattern) && namePattern.test(config.pattern)) {
     config.longer_alt = NAME;
+    // 'application' IS-A KEYWORD which in turn IS-A NAME --> all keywords are now valid identifiers.
+    config.parent = KEYWORD;
```
MathieuAA: > all keywords are now valid identifiers.

I'm not sure I understand this statement.

bd82: it is the inheritance chain of the tokens used as classifications

An "APPLICATION" extends a "KEYWORD" which in turn extends "NAME".
so this means that an "APPLICATION" is **transitively** a "NAME"

The result of this is that:
```javascript
CONSUME(t.NAME); 
```
Will work on an "APPLICATION" token as it would on all KEYWORDS tokens.

MathieuAA: I understand the code, not the comment itself...

bd82: The full comment should be:
"all keywords **can now be used** as valid identifiers."

Are you seeking an explanation why this is compatible with the way pegjs did things?
And why the default behavior for Chevrotain is different?

MathieuAA: No I understand, the comment can be misleading that's all.
Can you remove it? the code is enough I think.

bd82: sure

bd82:
Removed confusing comment.

PR #159, opened 2017-10-11
Assortment of fixes

bd82:
* Fixed MIN_MAX keywords parent.
* Consistent naming for grammar rules.
* Removed a ".only" mocha suffix and added an eslint rule to detect such
  errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "eslint": "4.8.0",
     "eslint-config-airbnb-base": "12.0.2",
     "eslint-plugin-import": "2.7.0",
+    "eslint-plugin-mocha": "^4.11.0",
     "istanbul": "0.4.5",
     "mocha": "4.0.1",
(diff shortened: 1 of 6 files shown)


bd82:
Important to merge this quick as only one test is running now due to mistake in previous PR. 😢

package.json
```diff
@@ -51,6 +51,7 @@
     "eslint": "4.8.0",
     "eslint-config-airbnb-base": "12.0.2",
     "eslint-plugin-import": "2.7.0",
+    "eslint-plugin-mocha": "^4.11.0",
```
deepu105: can you use a fixed version for this in the next PR? i'll merge this now

MathieuAA: @bd82 don't forget to run `yarn` to regenerate the yarn.lock file!

bd82: Opps, I used npm install instead of yarn add :(

PR #161, opened 2017-10-11
Compatibility test harness for parser migration.

bd82:
* It is skipped as it is meant to be run locally only.
* Allows checking syntax compatibility before the AST building is implemented.
* Currently 12/26 are failing, will examine those and will add commits to this PR.

### DIFF
--- a/test/spec/grammar/compatibility_test.js
+++ b/test/spec/grammar/compatibility_test.js
@@ -8,9 +8,9 @@ const _ = require('lodash');
 describe('Chevrotain Migration compatibility harness', () => {
   describe('Can Parse all the existing JDL samples in the test folder', () => {
     const samplesDir = path.resolve(__dirname, '../../test_files/');
-    const jdlSamples = _.filter(fs.readdirSync(samplesDir), file => _.endsWith(file, '.jdl'));
+    const jdlSamples = fs.readdirSync(samplesDir).filter(file => _.endsWith(file, '.jdl'));
 
-    _.forEach(jdlSamples, (currSample) => {
+    jdlSamples.forEach((currSample) => {
       // uncomment the "skip" to run the compatibility tests locally.
       it.skip(`Sample: '${currSample}'`, () => {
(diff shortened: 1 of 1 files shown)


test/spec/grammar/compatibility_test.js
```diff
@@ -0,0 +1,22 @@
...
+const fs = require('fs');
+const path = require('path');
+const _ = require('lodash');
+
+describe('Chevrotain Migration compatibility harness', () => {
+  describe('Can Parse all the existing JDL samples in the test folder', () => {
+    const samplesDir = path.resolve(__dirname, '../../test_files/');
+    const jdlSamples = _.filter(fs.readdirSync(samplesDir), file => _.endsWith(file, '.jdl'));
+
+    _.forEach(jdlSamples, (currSample) => {
```
MathieuAA: you needn't use lodash for looping

MathieuAA:
Not all files are meant to be parsed (some are just wrong). I'll filter them out later.

test/spec/grammar/compatibility_test.js
```diff
@@ -0,0 +1,22 @@
...
+const expect = require('chai').expect;
+const parse = require('../../../lib/dsl/poc/api').parse;
+const fs = require('fs');
+const path = require('path');
+const _ = require('lodash');
+
+describe('Chevrotain Migration compatibility harness', () => {
+  describe('Can Parse all the existing JDL samples in the test folder', () => {
+    const samplesDir = path.resolve(__dirname, '../../test_files/');
+    const jdlSamples = _.filter(fs.readdirSync(samplesDir), file => _.endsWith(file, '.jdl'));
```
MathieuAA: you needn't use lodash for filtering (it's okay for `endsWith` however)

test/spec/grammar/compatibility_test.js
deepu105: I will also prefer the native forEach

bd82: I will modify it.

Note that it does not really matter as this file is probably a **throwaway** file.
Once the migration is complete these JDL samples will be read as part of their own full test suites.

PR #182, opened 2018-02-23
Chevrotain migration

bd82:


bd82:
oy this has too many commits trying to resolve

PR #186, opened 2018-02-26
WIP: Added reserved keywords for Kotlin

thodorisbais:
Hello, 

while working on migrating the [`spring-controller`](https://github.com/jhipster/jhipster-kotlin/pull/16) for Kotlin, I noticed [this](https://github.com/thodorisbais/jhipster-kotlin/blob/985124019545d12d53cc324d96faba3ac36612f9/generators/spring-controller/prompts.js#L45) line and thought about enhancing the reserved keywords list by also adding Kotlin related references.

If you are fine with that, the convention used here is only about the Hard keywords Kotlin docs define.

Furthermore, there is something going wrong [here](https://github.com/thodorisbais/jhipster-core/blob/cfa2e8eca763664f3fb601eb81c84edd24f164ad/test/spec/core/jhipster/reserved_keywords_test.js#L87). 

If I understand correctly, the above test should pass.

Being in the need of feedback, I prefix this PR with a `WIP`.

Regards,
TB

### DIFF
--- a/lib/core/jhipster/reserved_keywords.js
+++ b/lib/core/jhipster/reserved_keywords.js
@@ -61,6 +61,11 @@ const ReservedWords = {
     'EXTENDS', 'INT', 'SHORT', 'TRY', 'CHAR', 'FINAL', 'INTERFACE', 'STATIC', 'VOID', 'CLASS', 'FINALLY',
     'LONG', 'STRICTFP', 'VOLATILE', 'CONST', 'FLOAT', 'NATIVE', 'SUPER', 'WHILE'
   ],
+  KOTLIN: [
+    'AS', 'BREAK', 'CLASS', 'CONTINUE', 'DO', 'ELSE', 'FALSE', 'FOR', 'FUN', 'IF', 'IN', 'INTERFACE',
+    'IS', 'NULL', 'OBJECT', 'PACKAGE', 'RETURN', 'SUPER', 'THIS', 'THROW', 'TRUE', 'TRY', 'TYPEALIAS',
+    'VAL', 'VAR', 'WHEN', 'WHILE'
+  ],
   TYPESCRIPT: [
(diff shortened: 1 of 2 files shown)


lib/core/jhipster/reserved_keywords.js
```diff
@@ -207,7 +212,7 @@ function isReservedTableName(keyword, databaseType) {
 }
 
 function isReservedFieldName(keyword) {
-  return isReserved(keyword, 'ANGULAR') || isReserved(keyword, 'TYPESCRIPT') || isReserved(keyword, 'JAVA');
+  return isReserved(keyword, 'ANGULAR') || isReserved(keyword, 'TYPESCRIPT') || isReserved(keyword, 'JAVA') || isReserved(keyword, 'KOTLIN');
```
deepu105: I dont think we should check every blueprint language supported in JH core, this should be handled by the blueprint itself else this defeats the purpose of having independent blueprint

MathieuAA:
@thodorisbais I'm improving every test file on the project so I'll have a look at this.

thodorisbais:
Cool @MathieuAA, I also made a few changes in here, you can have it as a starting point and let me know if you need any help.
So, @deepu105 , if I understand correctly, you prefer having this functionality inside the blueprint itself,  right?

deepu105:
@thodorisbais yes, we cant keep adding checks or reserved keywords for every blueprint as it will make a tight coupling between them. I want to avoid that at any cost

thodorisbais:
@deepu105 fair enough, it's clear now. I haven't taken any close look at the jhipster projects before, I went directly to the jhipster-kotlin to learn kotlin and jhipster at once :joy:

deepu105:
Thats a wise choice :)


Thanks & Regards,
Deepu

MathieuAA:
@deepu105 can we close this?

PR #192, opened 2018-03-19
Improve Syntax Diagrams readability.

bd82:
These are the full diagrams using the labels.
https://rawgit.com/bd82/jhipster-core/labels/diagrams.html

And here is a screenshot showing the labels (quoted literals) used for some terminals.
screen shot 2018-03-19 at 9 32 41 pm

### DIFF
--- a/lib/dsl/lexer.js
+++ b/lib/dsl/lexer.js
@@ -39,6 +39,11 @@ function createToken(config) {
     }
   }
 
+  // readable labels for diagrams
+  if (_.isString(config.pattern) && !config.label) {
+    config.label = `'${config.pattern}'`;
+  }
+
   // concisely collects all tokens to be exported
   const newToken = chevrotain.createToken(config);
   tokens[config.name] = newToken;
--- a/lib/dsl/gen/generated_serialized_grammar.js
+++ b/lib/dsl/gen/generated_serialized_grammar.js
@@ -1,1653 +1,1653 @@
-/*eslint-disable */
+/* eslint-disable no-unused-vars */
 const serializedGrammar = [
-	{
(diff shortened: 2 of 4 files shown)


lib/dsl/lexer.js
```diff
@@ -39,6 +39,11 @@ function createToken(config) {
     }
   }
 
+  // readable labels for diagrams
+  if (_.isString(config.pattern) && config.label === undefined) {
```
MathieuAA: Checking for type in JS is something I avoid. Here, you check the pattern is a string (so it could be something else).
`config` looks like a custom object.
If it actually is, this is the form this object should have:
```javascript
config = {
  label: '',
  pattern: ...,
  type: ...
}
```
Or create a class and don't deal with this yourself :)

About `config.label`, the recommended check is `!config.label`.

MathieuAA:
The generated file should be formatted (double-quotes, correct indent).

lib/dsl/lexer.js
bd82: I don't understand the alternative you are suggesting, 
* config is a configuration object pattern for the createToken chevrotain API.
* I want to define an easy to read diagrams labels **automatically** for all simple tokens
* Simple tokens are those tokens whose pattern is a simple string instead of a regExp/Function.

The other alternative is to manually write the label for each token but that is more error prone.

I can modify the check to !config.label and check running the eslint --fix on the generated file.

bd82:
I fixed the conditional and the linting of the generated file.
I am uncertain what is supposed to be done about the runtime type checks
and/or runtime transformation of the tokenType configuration object as I don't yet understand the issue.

PR #199, opened 2018-03-21
Some Requested Style Changes.

bd82:


lib/dsl/validator.js
```diff
@@ -256,7 +259,7 @@ class JDLSyntaxValidatorVisitor extends BaseJDLCSTVisitorWithDefaults {
   checkConfigPropSyntax(key, value) {
     const propertyName = key.tokenType.tokenName;
     const validation = configPropsValidations[propertyName];
-    if (_.isUndefined(validation)) {
+    if (validation === undefined) {
```
bd82: I meant to check for undefined here, not null, empty strings, zero ...

MathieuAA: `!validation`

MathieuAA: It depends on what `configPropsValidations` returns. It's a custom function, so we know and can control what it returns.

bd82: fixed

PR #202, opened 2018-03-25
Check invalid flows to hit 100% coverage for dsl/api.js

bd82:


test/spec/grammar/dsl_api_test.js
```diff
@@ -0,0 +1,140 @@
+const expect = require('chai').expect;
+const tokens = require('../../../lib/dsl/lexer').tokens;
+const getSyntacticAutoCompleteSuggestions = require('../../../lib/dsl/api')
```
MathieuAA: You can ignore the next two lines breaks...

test/spec/grammar/dsl_api_test.js
```diff
@@ -0,0 +1,140 @@
...
+  .getSyntacticAutoCompleteSuggestions;
+const parse = require('../../../lib/dsl/api')
+  .parse;
+
+
+describe('JDL DSL API', () => {
+  describe('when wanting an AST', () => {
+    describe('with a valid input', () => {
+      it('returns an AST', () => {
+        const ast = parse(`
```
MathieuAA: Instead of using the string templating, you can just do:
```javascript
parse('entity A { field String }');
```

test/spec/grammar/dsl_api_test.js
```diff
@@ -0,0 +1,140 @@
...
+          javadoc: null
+        });
+      });
+    });
+
+    describe('with a lexing error', () => {
+      it('throws an error', () => {
+        const invalidInput = 'entity ± {';
+        expect(() => parse(invalidInput)).to
+          .throw('unexpected character: ->±<- at offset: 7, skipped 1 characters.');
```
MathieuAA: If you want to test that I suggest having more than one `it`:
  - `throws an error` => to test there's an error
  - `displays the illegal character` => to test the illegal character is thrown back at us (which is **nice** by the way)
  - `displays column and line information` => to test there's some kind of info about that

If you decide to change the API for whatever reason, I want my test to fail and yell at me that I should read the doc/changelog to be up-to-date.

test/spec/grammar/dsl_api_test.js
```diff
@@ -0,0 +1,140 @@
...
+    });
+  });
+
+  describe('when wanting an auto-completion', () => {
+    describe('with an empty text', () => {
+      it('provides suggestions', () => {
+        const input = '';
+        const result = getSyntacticAutoCompleteSuggestions(input);
+        expect(result).to.have.lengthOf(17);
+        expect(result).to.have.members([
```
MathieuAA: For L.79 & L80: consider replacing them by `deep.equal`. I don't know if it's applicable here, just suggesting.

test/spec/grammar/dsl_api_test.js
```diff
@@ -0,0 +1,140 @@
...
+      });
+    });
+    describe('with a custom start rule', () => {
+      it('provides suggestions', () => {
+        const input = 'lastName string ';
+        const result = getSyntacticAutoCompleteSuggestions(
+          input,
+          'fieldDeclaration'
+        );
+        expect(result).to.have.lengthOf(4);
```
MathieuAA: Same comment

test/spec/grammar/dsl_api_test.js
bd82: That is a good suggestion, but that is not part of code of this commit. (renamed file)

test/spec/grammar/dsl_api_test.js
bd82: That is a good suggestion, but that is not part of code of this commit. (renamed file)

test/spec/grammar/dsl_api_test.js
bd82: I just want to see that it throws "something" I will make it more general.

test/spec/grammar/dsl_api_test.js
bd82: I tested it, it won't work because deep.equal tests the order of elements as well
and when building the expected list of token via the grammar
my brain and the algorithm used in Chevrotain output a different order of elements 😄

bd82:
I used **regExps** to be **less** specific in testing the error messages to reduce the  potential breakage due to valid future changes in the source code.

MathieuAA:
I feel it's a bit worse. The tests should be easy to read.

bd82:
What about latest commit syntax with **multiple** assertions?
I've checked the error messages to be clear and easy to understand in case of failures.

test/spec/grammar/dsl_api_test.js
```diff
@@ -0,0 +1,146 @@
+const expect = require('chai').expect;
+const tokens = require('../../../lib/dsl/lexer').tokens;
+// This long line demonstrates why destructuring imports are better :)
```
MathieuAA: Nope. Please remove the line :)

bd82:
I split up the assertions and added a before clause when there are multiple assertions for the same "topic".

 **I do agree that it is better this way**, although I think getting a higher code coverage is more important (at this time) than creating "perfect" tests
*  (I have less time to invest in JHipster in the near future...)

bd82:
Due to this being a large migration I'd feel safer just by having high coverage even if
no tests (assertions) are included. Basically I'd like to avoid errors such as "undefined is not an object"
happening for consumers of this library.

Although the situation may already be better as the previous generated parser was excluded from the code coverage statistics...

Issue #198, opened 2018-03-21
Update the chevrotain parsing system for application parsing


MathieuAA:
Some options have been removed (`useCompass` for instance, and other were added).

MathieuAA:
@bd82 I'll be honest: it's a chore. It was painful to do it last week-end and I don't expect it to be fun now... I'll try to do it again for the new system.

bd82:
it should be very easy because they are not listed individually in the grammar anymore.

Steps to add a config prop:
1. Define token that belong to CONFIG_KEY category -  [Code](https://github.com/jhipster/jhipster-core/blob/chevrotain-migration/lib/dsl/lexer.js#L125-L129)
2. Define its type validations - [Code](https://github.com/jhipster/jhipster-core/blob/chevrotain-migration/lib/dsl/validator.js#L78-L82)
3. Winner winner chicken dinner.

PR #227, opened 2018-06-23
Align config keys with latest pegjs grammar.

bd82:

### DIFF
--- a/lib/dsl/validator.js
+++ b/lib/dsl/validator.js
@@ -40,7 +40,7 @@ const typeNamePattern = /^[A-Z][A-Za-z0-9]*$/;
 const enumNamePattern = /^[A-Z][A-Za-z0-9]*$/;
 const enumPropNamePattern = /^[A-Z][A-Za-z0-9_]*$/;
 const methodNamePattern = /^[A-Za-z][A-Za-z0-9-]*$/;
-const jhiPrefixNamePattern = /^[A-Za-z][A-Za-z0-9-]*$/;
+const jhiPrefixNamePattern = /^[A-Za-z][A-Za-z0-9-_]*$/;
 const alphabetic = /^[A-Za-z]+$/;
 const alphabeticLower = /^[a-z]+$/;
 const alphanumeric = /^[A-Za-z][A-Za-z0-9]*$/;
@@ -70,16 +70,11 @@ const configPropsValidations = {
     pattern: alphanumeric,
     msg: 'cacheProvider property'
   },
-  // TODO: this property name and value changed
-  HIBERNATE_CACHE: {
-    type: 'NAME',
-    pattern: alphabetic,
-    msg: 'hibernateCache property'
+  ENABLE_HIBERNATE_CACHE: {
+    type: 'BOOLEAN'
   },
-  CLUSTERED_HTTP_SESSION: {
-    type: 'NAME',
-    pattern: alphabetic,
-    msg: 'clusteredHttpSession property'
+  ENABLE_SWAGGER_CODEGEN: {
+    type: 'BOOLEAN'
   },
   WEBSOCKET: {
     type: 'NAME',
@@ -101,7 +96,6 @@ const configPropsValidations = {
     pattern: alphanumeric,
     msg: 'prodDatabaseType property'
   },
-  USE_COMPASS: { type: 'BOOLEAN' },
   BUILD_TOOL: {
     type: 'NAME',
     pattern: alphanumeric,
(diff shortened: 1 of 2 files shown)


lib/dsl/validator.js
```diff
@@ -168,7 +158,9 @@ const configPropsValidations = {
   },
   SKIP_USER_MANAGEMENT: { type: 'BOOLEAN' },
   SKIP_CLIENT: { type: 'BOOLEAN' },
-  SKIP_SERVER: { type: 'BOOLEAN' }
+  SKIP_SERVER: { type: 'BOOLEAN' },
+  UAA_BASE_NAME: { type: 'STRING' },
+  JHIPSTER_VERSION: { type: 'INTEGER' }
```
MathieuAA: Not sure, it could be `5.0.0`, which is a string. Or `5.0.0-beta3`

deepu105: its definitely string

bd82: Changed it to string. Note that the pegjs version only supported x.y.z syntax:
```
5.0.0
5.2.4
```

But now there is also support for the suffix part of the semantic version.
At the cost of wrapping this in an actual string literal with its delimiters.
```
"5.0.0"
"5.2.4"
"6.0.0-beta"
```

bd82:
I think this PR may resolve #198 
?

lib/dsl/validator.js
MathieuAA: Yeah, this was a shortcoming of mine not thinking about a possible "beta", or "alpha", etc.
About the quotes, if we add them here we must add them everywhere, and I'm against that. Can we do without the quotes?

MathieuAA: I don't see the quotes element in your PR, am I missing something?

bd82: > I don't see the quotes element in your PR, am I missing something?

The application config is implemented in a generic manner.
So from a grammar perspective every config key could have any potential value.
https://github.com/jhipster/jhipster-core/blob/master/lib/dsl/jdl_parser.js#L412-L430

The validator.js file adds constraints **post** parsing to create syntax errors if and when
wrong types of values have been provided for specific keys.

This makes it very easy to customize the possible configuration keys and values
without an ever increasing grammar file. So in this specific case there was no need to change the grammar, only the constraints on the grammar.

bd82: >  About the quotes, if we add them here we must add them everywhere, and I'm against that. Can we do without the quotes?

When you say everywhere, where else do you mean?
Somewhere that only accepts a alphaNumeric name for example would not require any quotes...

It may be possible to accept this without quotes, but it would be risky as it could conflict
with other kinds of tokens,
e.g:
- is "5.2" a decimal number or a version number? 
- is "5.2-beta" a semantic version token or an arithmetic expression "5.2 - beta" (beta is a variable)

 and if we do this at the parser level it would allow whitespace between the parts e.g:
```
5 . 0 .  
3     -  beta
```

bd82: I think that quotes are only needed when a simple NAME/INTEGER token or combinations of will not suffice.
Just like in JavaScript.

```javascript
myObj.foo = 5 // accessing member using identifier
myObj.["5.0.2-beta"] = 5 // accessing member using string literal and square blocks because '5.0.2-beta" is not a valid javascript identifier.
```

MathieuAA: Well, aren't we the only ones saying that whatever is after `jhipsterVersion` is a string having a finite form like `[0-9]+\.[0-9]+\.[0-9]+(-[a-z]+)?`?
The JDL won't have variables, ever. This is a 100% certainty. And I can't think of a valid use case where having something like `5.2` would be good for the JDL. This is not a programming language.

bd82: In that case we combine these 5-7 tokens to a single parser rule to designate a version.
It would still allow whitespace between the different rules as our parser is whitespace insensitive.
But that is a minor concern (a formatter can fix that).

I will try to implement this later today or tomorrow.

MathieuAA: > It would still allow whitespace between the different rules as our parser is whitespace insensitive.

How so?

@bd82 take your time, I'm trying to get annotations in with the new system

bd82: O.k. so I'm implementing a semVer **literal** for the jHipsterVersion config key.
And it looks like a simple syntax is easy to implement.

But how **complete** do you expect the solution to be?
The the full specs here:
* https://docs.npmjs.com/misc/semver
* Just the [range syntax](https://docs.npmjs.com/misc/semver#range-grammar) is a full mini grammar by itself.
* tilde / caret modifiers
* Partial numbers ("3.2") is also a semantic version.
* 'X' may be used as a wild card.

Implementing a full parser for semVer is way outside the scope....

So do you prefer limiting the syntax of semVer variants allowed to avoid quotes
and thus potentially limiting the **functionality** of that config property.
Or do you prefer limiting the syntax of the JDL (forcing quotes).
and enabling all functionality of semVer versions?

bd82: For reference here is a full semVer parser
https://github.com/steveklabnik/semver-parser/tree/master/src

It probably does more than we need, but still provides a feeling of scope.

PR #228, opened 2018-06-24
Fixes for windows

bd82:


bd82:
I don't know if this solves a runtime production bug or not.
But I do know that production code threw a bug during tests on windows so that is likely.

package.json
```diff
@@ -43,19 +43,20 @@
     }
   ],
   "dependencies": {
+    "chevrotain": "3.6.1",
+    "fs-extra": "^6.0.1",
```
MathieuAA: Versions should be fixed

lib/utils/file_utils.js
```diff
@@ -17,8 +17,7 @@
  * limitations under the License.
  */
 
-const fs = require('fs');
-const path = require('path');
+const fs = require('fs-extra');
```
MathieuAA: I'd rather use the `fse` name, so as not get confused between fs and fs-extra.

bd82:
> I don't understand. What bug?

```
"before all" hook failed

Error: ENOENT: no such file or directory, mkdir 'toto\titi\tutu'
    at Object.fs.mkdirSync (fs.js:874:3)
    at folderStructure.reduce (lib\utils\file_utils.js:67:10)
    at Array.reduce (<anonymous>)
    at Object.createDirectory (lib\utils\file_utils.js:63:19)
    at Context.before (test\spec\utils\files_utils_test.js:98:19)
```

lib/utils/file_utils.js
bd82: See: https://github.com/jprichardson/node-fs-extra#usage

> fs-extra is a drop in replacement for native fs. All methods in fs are attached to fs-extra. All fs methods return promises if the callback isn't passed.
You don't ever need to include the original fs module again:

bd82: we can still use a separate name if you prefer though...

MathieuAA: I've read that, thanks :) That's why I want the `fse` name.

bd82: sure

bd82:
There are three commits here so please **squash**  😄

package.json
bd82: fixed

MathieuAA:
Thank you!!

bd82:
No problem 😄

PR #229, opened 2018-06-24
Less verbose style to write the parser.

bd82:


bd82:
This is a subjective suggestion

bd82:
Sure just a suggestion

Issue #234, opened 2018-06-25
Bundling for use in the JDL Studio


bd82:
Is this package ever bundled (webpack?) for use in the browser?
If so special care must be taken when bundling Chevrotain grammars.

http://sap.github.io/chevrotain/docs/FAQ.html#WEBPACK

deepu105:
The JDL studio uses the output from the PegJS parser https://github.com/jhipster/jdl-studio/blob/gh-pages/bower.json#L21

So we would need something similar from this one as well to use in JDL studio. 

I totally forgot about this one

bd82:
So we have the following options:

1. Create a Bundled version as part of the jhipster-core npm package.
   * With safe name mangling.
2. Perform safe bundling as part of the JDL Studio project.
   * Feels like a mistake, I am not sure that it should be the studio's concern.
3. Apply grammar [serialization and de-serialization](https://github.com/SAP/chevrotain/tree/master/examples/parser/serialized_grammar) to make the parser insensitive to bundling transformations.

There is a secondary issue that the studio uses bower to manage dependencies so if we need to consume generated artifacts (e.g: jhipster_core.**min.js**) it won't work unless we add those generated files to the source control.
Perhaps it is time for the studio to consume dependencies via npm?

bd82:
@deepu105 @MathieuAA 

How should we proceed here?

deepu105:
Definitely not option 2.

I will prefer option 1, but then it might be tricky as jhipster-core is a nodejs project and packaging the entire thing and importing in a web app will not work unless we use something like browserify.

About the bower issue, one reason I didn't migrate it was the same issue as above I didn't want the entire package, but just a particular file to deal with the grammar validation & parsing.

Another option is to provide the validation & parsing as a REST service from JHipster online.

Another option is to use jdl-ide for the validations and parsing instead of jhipster-core
I'm pinging @jbadeau @colameo team as well for their opinion

bd82:
One option would be to only bundle the [DSL](https://github.com/jhipster/jhipster-core/tree/master/lib/dsl) directory as part of the npm package of JHipster-Core.

That bit only depends on lodash and will provide you with similar functionality (but different new APIs)

deepu105:
That could also work. JDL studio only needs the DSL parser (the one PegJS was providing)

bd82:
@deepu105 I think #240 is complete.

You should probably build the bundle locally (yarn bundle)
and experiment with it.

[...]

PR #240, opened 2018-06-30
Bundle DSL related logic for JDL studio.

bd82:
Hi @MathieuAA, @deepu105 

Please review this, if you like this approach I will finish the required smoke testing.

Thing to note:
* lodash is not bundled here instead users of this library will have to include lodash in their env.
* Chevrotain is bundled because there will be a stronger coupling between a chevrotain version
  and the DSL.
* We need to consider if we want to run babel on the output, otherwise it won't work on IE11
   - But IE11 is not really a browser 😄 
* We may need babel anyhow due to different EcmaScript syntax levels supported in different browsers (even ever green ones). but maybe that is not needed as JHIpster-core targets a slightly older version of V8 (on node.js 8) so it "should'" also work on more modern JS runtimes.
  - Should and Will are different things though 😄


bd82:
related to #234

package.json
```diff
@@ -59,11 +61,19 @@
     "nyc": "12.0.2",
     "pegjs": "0.10.0",
     "sinon": "6.0.0",
-    "sinon-chai": "3.2.0"
+    "sinon-chai": "3.2.0",
+    "uglifyjs-webpack-plugin": "^1.2.7",
```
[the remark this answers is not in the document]
bd82: thanks, ```yarn add``` always adds ranges and I don't notice it 😄

webpack.config.js
```diff
@@ -0,0 +1,41 @@
+const _ = require('lodash');
+const path = require('path');
+const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
+const { tokens } = require('./lib/dsl/lexer');
+
+const allTokenTypeNames = _.mapValues(tokens, tokenType => tokenType.name);
+
+module.exports = {
+  mode: 'production',
+  entry: './lib/dsl/exports.js',
```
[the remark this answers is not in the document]
bd82: I am pretty sure webpack handles this automatically, so it would be **redundant**...
See their docs:
https://webpack.js.org/concepts/entry-points/#src/components/Sidebar/Sidebar.jsx

I am using both Mac and Windows for development and would have encountered such 
an issue if it existed.

webpack.config.js
```diff
@@ -0,0 +1,41 @@
...
+  output: {
+    path: path.resolve(__dirname, './dist/'),
```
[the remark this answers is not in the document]
bd82: I will check on my windows machine, again I think webpack is smart enough here 😄

test/spec/bundle/bundle_test.js
```diff
@@ -0,0 +1,34 @@
+const expect = require('chai').expect;
+const _ = require('lodash');
+const distBundle = require('../../../dist/jhipster.min');
+/**
+ * Smoke Test for the Grammar UMD bundle.
+ */
+describe('JDL DSL UMD bundle', () => {
```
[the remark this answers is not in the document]
bd82: It is a **good question** about the need for the tests.
Some kind of minimal smoke tests will help us be "secure" the bundling process works as expected. but on the other hand we are not really testing in a browser so the value is reduced.
Also these are not fully official APIs, more like semi official for a single use case of the JDL studio.

**So maybe the value of these tests is not very large**? and @deepu105 will just "scream"
If the JDL studio breaks when he attempts to upgrade the jhipster-core version used?

In Chevrotain I run browsers tests using sauce labs on multiple browser versions
and testing both the minifed and unminifed bundle. But these are official APIs I am testing
and it took quite a while to set this up and even still its a little unstable...

lib/dsl/exports.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * This file defines the exported JDL DSL related APIs
+ * This file will be used as webpack entry point to produce a UMD bundle.
+ */
+
+const tokens = require('./lexer').tokens;
```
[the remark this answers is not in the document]
bd82: will fix

bd82:
What about using Babel to transpile to ES5?
 does the JDL-Studio work under IE11?

webpack.config.js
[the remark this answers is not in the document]
deepu105: Yes webpack will handle all such stuff

webpack.config.js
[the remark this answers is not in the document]
deepu105: Yes I'm pretty sure webpack handles this. I too would have noticed otherwise

deepu105:
@bd82 I never tested it in IE. I'm ok if it works only for modern browsers. Wouldn't want to complicate things just for IE

webpack.config.js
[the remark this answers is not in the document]
bd82: Relevant docs:
https://webpack.js.org/configuration/watch/#windows-paths

webpack.config.js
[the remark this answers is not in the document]
bd82: Relevant docs: https://webpack.js.org/configuration/watch/#windows-paths

bd82:
* Removed automated node.js tests.
* Added **manual** mocha html test runner.
* Tested on chrome/firefox/safari.

lib/dsl/exports.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * This file defines the exported JDL DSL related APIs
+ * This file will be used as webpack entry point to produce a UMD bundle.
+ */
+const lexer = require('./lexer');
+const api = require('./api');
+const JDLParser = require('./jdl_parser');
+
+const tokens = lexer.tokens;
```
deepu105: cant we just deconstruct these? `const { tokens, JDLLexer } = require('./lexer')` and same for api

webpack.config.js
```diff
@@ -0,0 +1,41 @@
...
+    filename: 'jhipster.min.js',
```
deepu105: can we call this file `jdl-core.min.js` please

webpack.config.js
```diff
@@ -0,0 +1,41 @@
...
+    library: 'jdl',
+    libraryTarget: 'umd',
+    // https://github.com/webpack/webpack/issues/6784
+    globalObject: 'typeof self !== \'undefined\' ? self : this'
+  },
+  externals: {
+    // https://webpack.js.org/guides/author-libraries/#externalize-lodash
+    lodash: {
+      commonjs: 'lodash',
```
deepu105: is lodash the only dependency that we are bundling?

test/bundle_test.html
```diff
@@ -0,0 +1,70 @@
...
+        let ast;
+
+        before(() => {
+          ast = parse('@service(serviceClass) entity A {field String}');
+        });
+
+        it('returns an AST', () => {
+          expect(ast.entities).to.have.lengthOf(1);
+          expect(ast.entities[0]).to.deep.eql({
+            name: 'A',
```
deepu105: is the output format same as that pegJS used to give or do we need to update studio to work with new format?

webpack.config.js
bd82: lodash is not bundled here, it is defined as external which means it is resolved at runtime.

test/bundle_test.html
bd82: The AST output is identical.

webpack.config.js
bd82: yes will fix

lib/dsl/exports.js
bd82: @MathieuAA does not like deconstruct syntax 😄

MathieuAA: Yeah I don't but if it's just for that let's go for it...

deepu105: @MathieuAA but why????????? deconstruction is so coool :smile_cat:

webpack.config.js
deepu105: Sorry, yes I know, wrong choice of word. what I meant was to ask if lodash was the only dependency and I checked the chevrotain package.json after writing this. Thanks anyway

test/bundle_test.html
deepu105: awesome

webpack.config.js
bd82: Chevrotain has a dependency to regexp-to-ast npm package (a regexp parser I wrote).
But I make sure to bundle it inside the chevrotain.min.js bundle for ease of consumption in the browser so you don't have to load anything extra except lodash.

You can see this in the sample html mocha runner test provided in this commit

lib/dsl/exports.js
bd82: EcmaScript modules:

```javascript
import {x, y, z} from "./bar"
export const foo = 666;
```

Are already supported in all ever green browsers.
I guess that in a year or two we will have it supported in all LTS versions of node.js
So we could use it directly in node packages without transpiration 😄

webpack.config.js
bd82: Actually it does not matter because Chevrotain is bundled inside the JDL-core.js bundle
so webpack handles the transitive dependencies here.

webpack.config.js
bd82: changed the file name to jdl-core.min.js **and** the library name
to match as "jdlCore".

lib/dsl/exports.js
bd82: I left it as is.

You can change this later to whatever you want in a **separate commit**.
I think consistency is important but alos think using newer syntax is important.
so I am **conflicted**

I suggest you make a decision if the upgrade the syntax in the entire project
or just wait until all jhipster supported versions would have the newest import/export syntax 😄

bd82:
@MathieuAA @deepu105   I think this can be merged now?

Issue #241, opened 2018-06-30
Potential grammar change to review


bd82:
In 3.02 many of the application config values were changed to be alpha numeric
instead of alphabetic.

See: https://github.com/jhipster/jhipster-core/commit/05a907747870a9d2d7a662c773752419d475e1e0

The Chevrotain implementation allows alpha numeric values
with the **single constraint** that they must start with a letter.

e.g:
```
application foo {
    config {
       // valid (at least syntax) in 3.02 
       databaseType 666
    }
}

application foo {
    config {
       // valid (at least syntax) in 3.02 
       databaseType 111mssql
    }
}
```

With the Chevrotain Parser the two examples above are invalid.
I am guessing that these examples (starting with number / number only) should be invalid.
Please review and let me if these keys should have values starting with numbers.

bd82:
@MathieuAA please have a look.

ruddell:
The only JHipster variables I'm aware of that have numbers are `oauth2` (`authenticationType`) and `h2Disk`/`h2Memory` (`devDatabaseType`).

Regarding variables set by the user, `baseName` and `uaaBaseName` can start with letters or numbers

MathieuAA:
@ruddell @bd82 They may be the only examples where there are letters & numbers, but we're not safe from tools having a number in their name... The safe option would be to support having them in the option names.
However, the chance of having a number first is pretty low, so we can have a letter first then (almost) whatever next.

bd82:
Well since having a number first was only supported in the most recent patch version
and the chance of valid values starting with numbers is small I suggest we will handle this
as a **patch** if and when this issue will be raised by users.

I can solve this mini ambiguity (numbers vs identifiers starting with number) but I'm not sure its worth dealing with it right now if it is mostly theoretical.

bd82:
If this approach is acceptable we can close this issue.

MathieuAA:
Well for me it 50/50... Let's fix it once and for all for the two properties so that we don't have to deal with this later.

bd82:
I will provide the "fix" in a PR
and you can decide if to merge it or not.

I'm just not fond of adding ambiguities (even simple resolvable ones)
When I'm not sure there is a real need.

bd82:
I just had another look at the two properties @ruddell mentioned: 
* baseName 
* uaaBaseName

MathieuAA:
[...]

PR #245, opened 2018-07-04
POC for digit as first char in baseName property value.

bd82:


bd82:
Discussion in #241

lib/dsl/ast_builder.js
```diff
@@ -482,6 +482,8 @@ class JDLAstBuilderVisitor extends BaseJDLCSTVisitor {
       return stringImage.substring(1, stringImage.length - 1);
     } if (ctx.BOOLEAN) {
       return ctx.BOOLEAN[0].image === 'true';
+    } if (ctx.NAME_CAN_START_DIGIT) {
```
MathieuAA: Didn't notice what was previously there, but each if should start on its own line.

lib/dsl/jdl_parser.js
```diff
@@ -571,6 +571,7 @@ module.exports = class JDLParser extends Parser {
   configValue() {
     this.RULE('configValue', () => {
       this.OR([
+        { ALT: () => this.CONSUME(LexerTokens.NAME_CAN_START_DIGIT) },
```
MathieuAA: I don't like it at all. There's no control over each property, we just say "okay, these types are possible and let the validator check everything's okay".

lib/dsl/lexer.js
```diff
@@ -23,8 +23,51 @@ const Lexer = chevrotain.Lexer;
 
 const tokens = {};
 
+const baseNamePattern = 'baseName';
```
MathieuAA: It's not really a pattern... its use is, but it's not. The name is kinda misleading.

lib/dsl/lexer.js
```diff
@@ -23,8 +23,51 @@ const Lexer = chevrotain.Lexer;
...
 const namePattern = /[a-zA-Z_][a-zA-Z_\-\d]*/;
-// All the names tokens of the pegjs implementation have been merged into
+// using start anchor because we are matching with this regexp directly in a custom pattern.
+const namePatternWithStartDigit = /^[\da-zA-Z_][a-zA-Z_\-\d]*/;
```
MathieuAA: Some rules should be enforced here, whether to use `\d` or `0-9`. Having both is not that pretty. By the way, why isn't it ` const namePatternWithStartDigit = /^[\da-zA-Z_][a-zA-Z_\-\d]*$/; ` the `$` sign marking the end of the local matching)

lib/dsl/lexer.js
```diff
@@ -23,8 +23,51 @@ const Lexer = chevrotain.Lexer;
...
+/**
+ * Accepts an array of array representing character ranges
+ * e.g: [['a', 'd'], ['0', '2']]
+ *
+ * And returns a set of the "expanded" ranges in charCodes values.
+ * e.g: [97, 99, 98 , 100, 48, 49, 50]
+ *   - 'a'.charCodeAt(0) === 97
+ *   - 'a'.charCodeAt(0) === 100
+ */
+function getRanges(charRanges) {
```
MathieuAA: nice idea

lib/dsl/lexer.js
```diff
@@ -23,8 +23,51 @@ const Lexer = chevrotain.Lexer;
...
+
+    ranges = ranges.concat(_.range(startCode, endCode));
+  });
+
+  return _.uniq(ranges);
+}
+
+// Special NAME Token which follows a baseName config value.
+// It allows starting with numbers or even being only digits.
+const NAME_CAN_START_DIGIT = chevrotain.createToken({
```
MathieuAA: some wording to do, "can start with digit", nothing critical but I get the meaning

lib/dsl/lexer.js
```diff
@@ -23,8 +23,51 @@ const Lexer = chevrotain.Lexer;
...
+  pattern: (text, offset, matchedTokens) => {
+    const lastToken = _.last(matchedTokens);
+    if (lastToken && lastToken.image === baseNamePattern) {
+      return namePatternWithStartDigit.exec(text.substring(offset));
+    }
+
+    return null;
+  },
+  // this optional property enables performance optimizations in the Lexer.
+  start_chars_hint: getRanges([['0', '9'], ['a', 'z'], ['A', 'Z'], ['_', '_']]),
```
MathieuAA: If there are inputs like `['_', '_']`, maybe not iterating over it is better by passing a string and checking with `Array::isArray`

lib/dsl/jdl_parser.js
bd82: That is basic separation of concerns, Syntax versus Type Checks.
Also implemented like  this in the JDL IDE.
https://github.com/jhipster/jhipster-ide/blob/master/jdl/io.github.jhipster.jdl/src/io/github/jhipster/jdl/JDL.xtext#L74

Its like applying a JSON schema on a JSON input.
You can expand it to the full grammar with the very specific types for each config key.
There is nothing preventing doing so. 

A gentle reminder that you yourself complained that maintaining the old system was a chore.
https://github.com/jhipster/jhipster-core/issues/198#issuecomment-374896968
This way its much easier to add new config keys (you don't even have to change the parser!).

lib/dsl/lexer.js
bd82: Yes I see your point

lib/dsl/lexer.js
bd82: I agree.

lib/dsl/lexer.js
bd82: Will fix if we move ahead with this change.

lib/dsl/lexer.js
bd82: I don't understand the comment can you demonstrate?

lib/dsl/ast_builder.js
bd82: It was a series of "else if" which you modified to plain ifs in this commit
https://github.com/jhipster/jhipster-core/commit/fe6614892f03cf02b9baf84a17546ea93140a605#diff-9fa3402fa25eaa022ea233eb8dbb86c9
Which is why the formatting is not valid...

But style X or style Y are irrelevant, the main issue is that formatting is not done **automatically.** Basically we(Humans) should not invest time in doing something a computer would do better and more consistently (prettier/standardjs).

Particularly as we all have day jobs and limited time to invest in OSS. 😭

deepu105: I'm always ok to add prettier. I thought it was already done here??

deepu105: I'm sorry just found the old ticket and realized I was being ignorant there, I think aging a bit more has made me a bit wiser :wink: 
https://github.com/jhipster/jhipster-core/issues/158

bd82: np, its good to use this example use case to re-raise the question of standard consistent formatting tool. 👍

lib/dsl/jdl_parser.js
MathieuAA: Well as Deepu said, aging made me a bit wiser ;)

lib/dsl/lexer.js
MathieuAA:
```javascript
getRanges([
  ['0', '9'],
  ['a', 'z'],
  ['A', 'Z'],
  '_'
]);
```

bd82: Yes this version is nicer.

PR #246, opened 2018-07-05
Example of allowing two "Types" for a config property.

bd82:
Related to discussion in #241


bd82:
This is not tested yet but this is the general concept.

PR #255, opened 2018-09-05
JDL generator has wrong regex validation for language(can't pass zh-cn,zh-tw)

ttoommbb:
The language regex can't contains dash character, we need to support that.
Accoridng to https://tools.ietf.org/html/rfc5646, some language code also
may contain number characters. 

=================

  676 passing (1s)

$ npm run coverage

> jhipster-core@3.3.2 coverage /home/bohong/workspace/jhipster-core
> nyc --reporter=text-summary report

=============================== Coverage summary ===============================
Statements   : 96.75% ( 2500/2584 )
Branches     : 93.05% ( 1125/1209 )
Functions    : 96.41% ( 671/696 )
Lines        : 96.82% ( 2470/2551 )
================================================================================
Done in 7.03s.

  - [✔️] [Travis tests](https://travis-ci.org/jhipster/jhipster-umk/pull_requests) are green
  - [✔️] Tests are added where necessary
  - [✔️] Documentation is added/updated where necessary
  - [✔️] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/jhipster-core/blob/master/CONTRIBUTING.md) are followed

### DIFF
--- a/lib/dsl/validator.js
+++ b/lib/dsl/validator.js
@@ -33,6 +33,7 @@ const ALPHABETIC = /^[A-Za-z]+$/;
 const ALPHABETIC_LOWER = /^[a-z]+$/;
 const ALPHANUMERIC = /^[A-Za-z][A-Za-z0-9]*$/;
 const ALPHANUMERIC_DASH = /^[A-Za-z][A-Za-z0-9-]*$/;
+const LANGUAGE_PATTERN = /^[a-z]+(-[A-Za-z0-9]+)*$/;
 
 const configPropsValidations = {
   APPLICATION_TYPE: {
@@ -96,7 +97,7 @@ const configPropsValidations = {
   },
   LANGUAGES: {
     type: 'list',
-    pattern: ALPHABETIC_LOWER,
+    pattern: LANGUAGE_PATTERN,
     msg: 'languages property'
   },
   MESSAGE_BROKER: {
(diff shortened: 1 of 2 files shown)


MathieuAA:
It should be working.... anyway, thanks for fixing it, I'll have a look at it today

lib/dsl/validator.js
```diff
@@ -33,6 +33,7 @@ const ALPHABETIC = /^[A-Za-z]+$/;
 const ALPHABETIC_LOWER = /^[a-z]+$/;
 const ALPHANUMERIC = /^[A-Za-z][A-Za-z0-9]*$/;
 const ALPHANUMERIC_DASH = /^[A-Za-z][A-Za-z0-9-]*$/;
+const LANGUAGE_PATTERN = /^[a-z]+(-[A-Za-z0-9]+)*$/;
```
MathieuAA: Should be `/^[a-z]+(-[A-Za-z0-9]+)?$/`, because we don't want something like `ab-cd-ef`

ttoommbb: I used to create a more complex regex to match more cases defined in rfc5646.
//https://tools.ietf.org/html/rfc5646#section-4.4.1
const LANGUAGE_PATTERN = /^[a-z]{2,8}(-[A-Za-z0-9]{1,9}){0,8}$/;

Which should be allowed to have languages ab-cd-ef, ref:
https://www.w3.org/International/articles/bcp47/

MathieuAA: I didn't know that, you're right! Thanks for pointing it out.

PR #269, opened 2018-10-09
Add support for alphanumeric package names

jsm174:
Example package names could now be microservice1, microservice2, etc.

Used 5.4.1 and the following JDL for testing:

```
application {
  config {
    baseName uaa,
    applicationType uaa,
    packageName com.example.uaa,
    authenticationType uaa,
    uaaBaseName "uaa",
    prodDatabaseType postgresql,
    serviceDiscoveryType eureka,
    buildTool gradle
  }
}

application {
  config {
    baseName gateway,
    applicationType gateway,
    authenticationType uaa,
    packageName com.example.gateway,
    uaaBaseName "uaa",
    databaseType no,
    serviceDiscoveryType eureka,
    clientFramework react,
    buildTool gradle
  }
  entities A, B, C, D
}

application {
  config {
    baseName microservice1,
    applicationType microservice,
    authenticationType uaa,
    packageName com.example.microservice1,
    uaaBaseName "uaa",
    prodDatabaseType postgresql,
    serviceDiscoveryType eureka,
    buildTool gradle
  }
  entities A, B
}

application {
  config {
    baseName microservice2,
    applicationType microservice,
    packageName com.example.microservice2,
    authenticationType uaa,
    uaaBaseName "uaa",
    prodDatabaseType postgresql,
    serviceDiscoveryType eureka,
    buildTool gradle
  }
  entities C, D
}

entity A {
    data String
}

entity B {
    data String
}

entity C {
    data String
}

entity D {
    data String
}

dto * with mapstruct
service * with serviceClass

microservice A, B with microservice1
microservice C, D with microservice2
```

  - [ ] [Travis tests](https://travis-ci.org/jhipster/jhipster-umk/pull_requests) are green
  - [] Tests are added where necessary

### DIFF
--- a/lib/dsl/validator.js
+++ b/lib/dsl/validator.js
@@ -33,6 +33,7 @@ const ALPHABETIC = /^[A-Za-z]+$/;
 const ALPHABETIC_LOWER = /^[a-z]+$/;
 const ALPHANUMERIC = /^[A-Za-z][A-Za-z0-9]*$/;
 const ALPHANUMERIC_DASH = /^[A-Za-z][A-Za-z0-9-]*$/;
+const ALPHANUMERIC_LOWER = /^[a-z][a-z0-9]*$/;
 const LANGUAGE_PATTERN = /^[a-z]+(-[A-Za-z0-9]+)*$/;
 
 const configPropsValidations = {
@@ -112,7 +113,7 @@ const configPropsValidations = {
   },
   PACKAGE_NAME: {
     type: 'qualifiedName',
-    pattern: ALPHABETIC_LOWER,
+    pattern: ALPHANUMERIC_LOWER,
     msg: 'packageName property'
   },
   PROD_DATABASE_TYPE: {
--- a/lib/dsl/validator.js
+++ b/lib/dsl/validator.js
@@ -29,11 +29,11 @@ const ENUM_NAME_PATTERN = /^[A-Z][A-Za-z0-9]*$/;
 const ENUM_PROP_NAME_PATTERN = /^[A-Z][A-Za-z0-9_]*$/;
 const METHOD_NAME_PATTERN = /^[A-Za-z][A-Za-z0-9-]*$/;
 const JHI_PREFIX_NAME_PATTERN = /^[A-Za-z][A-Za-z0-9-_]*$/;
+const PACKAGE_NAME_PATTERN = /^[a-z_][a-z0-9_]*$/;
 const ALPHABETIC = /^[A-Za-z]+$/;
 const ALPHABETIC_LOWER = /^[a-z]+$/;
 const ALPHANUMERIC = /^[A-Za-z][A-Za-z0-9]*$/;
 const ALPHANUMERIC_DASH = /^[A-Za-z][A-Za-z0-9-]*$/;
-const ALPHANUMERIC_LOWER = /^[a-z][a-z0-9]*$/;
 const LANGUAGE_PATTERN = /^[a-z]+(-[A-Za-z0-9]+)*$/;
 
 const configPropsValidations = {
@@ -113,7 +113,7 @@ const configPropsValidations = {
   },
   PACKAGE_NAME: {
     type: 'qualifiedName',
-    pattern: ALPHANUMERIC_LOWER,
+    pattern: PACKAGE_NAME_PATTERN,
     msg: 'packageName property'
   },
   PROD_DATABASE_TYPE: {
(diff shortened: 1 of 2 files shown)


ruddell:
The regex in the generator [is slightly different](https://github.com/jhipster/generator-jhipster/blob/b46c92b0c58c4e893b99679c0aed0e7e795b3602/generators/server/prompts.js#L60-L63), I think we should match that to prevent invalid package names.  It looks like they can't start with numbers

` /^([a-z_]{1}[a-z0-9_]*(\.[a-z_]{1}[a-z0-9_]*)*)$/.test(input)`

jsm174:
The regex in lib/dsl/validator.js was:

`/^[a-z]+$/`

That doesn't even look for periods. Is somewhere else parsing each section? 

By changing it to 

`/^[a-z][a-z0-9]*$/`

we are still requiring the first character to be a - z.

ruddell:
The regex I posted above ensures the first character after a period is a letter, which is required by a Java package name.  

```
// invalid package
package com.example.2gateway;

// valid package
package com.example.gateway;

```

jsm174:
I agree that the regex you are suggesting is more correct, but I think the JDL parser is not working like you are assuming:

```
application {
  config {
    baseName microservice2,
    applicationType microservice,
    packageName com.example.2microservice,
    authenticationType uaa,
    uaaBaseName "uaa",
    prodDatabaseType postgresql,
    serviceDiscoveryType eureka,
    buildTool gradle
  }
  entities C, D
}
```

The fix I proposed will not allow `com.example.2microservice`:

```
jhipster import-jdl app.jdl
Using JHipster version installed globally
Executing import-jdl app.jdl
Options: from-cli: true
The JDL is being parsed.
error: Error message:
	MismatchedTokenException: Expecting token of type --> NAME <-- but found --> '2' <--
	at line: 47, column: 29
SyntaxError: Error: MismatchedTokenException: Expecting token of type --> NAME <-- but found --> '2' <--
	at line: 47, column: 29
ERROR! Error while parsing applications and entities from the JDL SyntaxError: Error: MismatchedTokenException: Expecting token of type --> NAME <-- but found --> '2' <--
	at line: 47, column: 29
```

You can test with:

```
npm uninstall -g generator-jhipster
npm install -g generator-jhipster@5.4.1
vi /usr/local/lib/node_modules/generator-jhipster/node_modules/jhipster-core/lib/dsl/validator.js
```

deepu105:
The . Seperator is split first and only the remaining tokens are validated
so no need to account for the period in validation, everything else can be
used in validation as in generator

lib/dsl/validator.js
```diff
@@ -33,6 +33,7 @@ const ALPHABETIC = /^[A-Za-z]+$/;
 const ALPHABETIC_LOWER = /^[a-z]+$/;
 const ALPHANUMERIC = /^[A-Za-z][A-Za-z0-9]*$/;
 const ALPHANUMERIC_DASH = /^[A-Za-z][A-Za-z0-9-]*$/;
+const ALPHANUMERIC_LOWER = /^[a-z][a-z0-9]*$/;
```
deepu105: Underscore also needs to be allowed so add a specific pattern here. PACKAGE_NAME

jsm174: Renamed ALPHANUMERIC_LOWER to PACKAGE_NAME_PATTERN and added underscores to regex in second commit.

lib/dsl/validator.js
```diff
@@ -112,7 +113,7 @@ const configPropsValidations = {
   },
   PACKAGE_NAME: {
     type: 'qualifiedName',
-    pattern: ALPHABETIC_LOWER,
+    pattern: PACKAGE_NAME_PATTERN,
```
MathieuAA: This basically means that a package could be like `_a._b._c`

jsm174: Correct.  

@ruddell regex from above has the ability to start with _. 

Then I went to:

https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html

> This can occur if the domain name contains a hyphen or other special character, if the package name begins with a digit or other character that is illegal to use as the beginning of a Java name, or if the package name contains a reserved Java keyword, such as "int". In this event, the suggested convention is to add an underscore

ruddell: Technically those work

deepu105: yup they are valid

deepu105:
thank you @jsm174

PR #273, opened 2018-10-16
Fix convertEntitiesToJDL when there is no relationship

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/8572

  - [ ] [Travis tests](https://travis-ci.org/jhipster/jhipster-umk/pull_requests) are green

### DIFF
--- a/lib/converters/json_to_jdl_entity_converter.js
+++ b/lib/converters/json_to_jdl_entity_converter.js
@@ -169,14 +169,15 @@ function addRelationshipsToJDL() {
 }
 
 function dealWithRelationships(relationships, entityName) {
-  if (relationships !== undefined) {
-    relationships.forEach(relationship => {
-      const jdlRelationship = getRelationship(relationship, entityName);
-      if (jdlRelationship) {
-        configuration.jdl.addRelationship(jdlRelationship);
-      }
-    });
+  if (!relationships) {
+    return;
   }
+  relationships.forEach(relationship => {
+    const jdlRelationship = getRelationship(relationship, entityName);
+    if (jdlRelationship) {
+      configuration.jdl.addRelationship(jdlRelationship);
+    }
+  });
 }
 
 function getRelationship(relationship, entityName) {
--- a/lib/converters/json_to_jdl_entity_converter.js
+++ b/lib/converters/json_to_jdl_entity_converter.js
@@ -169,12 +169,14 @@ function addRelationshipsToJDL() {
 }
 
 function dealWithRelationships(relationships, entityName) {
-  relationships.forEach(relationship => {
-    const jdlRelationship = getRelationship(relationship, entityName);
-    if (jdlRelationship) {
-      configuration.jdl.addRelationship(jdlRelationship);
-    }
-  });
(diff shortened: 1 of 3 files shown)


pascalgrimaud:
After launching `npm install`, a file is modified: `dist/jdl-core.min.js`
@MathieuAA : can you tell me if I should commit it in this PR ? If not, in that case, this file should be in git ignore ?

lib/converters/json_to_jdl_entity_converter.js
```diff
@@ -169,12 +169,14 @@ function addRelationshipsToJDL() {
 }
 
 function dealWithRelationships(relationships, entityName) {
-  relationships.forEach(relationship => {
-    const jdlRelationship = getRelationship(relationship, entityName);
-    if (jdlRelationship) {
-      configuration.jdl.addRelationship(jdlRelationship);
-    }
-  });
+  if (relationships !== undefined) {
```
MathieuAA: Clearer: 
```javascript
if (!relationships) {
  return;
}
```

MathieuAA:
You can forget about this file for now, It's only useful to commit it before releases at least.

pascalgrimaud:
nice to know, thanks Mathieu

test/spec/converters/json_to_jdl_entity_converter_test.js
```diff
@@ -251,6 +251,24 @@ describe('JSONToJDLEntityConverter', () => {
       });
     });
   });
+
+  describe('::convertEntitiesToJDL without relationship', () => {
```
MathieuAA: It should be a `context` case inside the `describe('::convertBlaBlaBla` up this file.
Like this:
```javascript
context('without relationships', () => {
  // your declarations & setup
  it('....', () => {

  });
});
```

I don't quite understand what you're testing here: that it doesn't fail or the fact that there's a table name?

test/spec/converters/json_to_jdl_entity_converter_test.js
```diff
@@ -251,6 +251,24 @@ describe('JSONToJDLEntityConverter', () => {
...
+    let jdlObject = null;
+
+    before(() => {
+      const entities = {
+        CassBankAccount: readJsonEntity('CassBankAccount')
+      };
+      // entities.Employee.relationships.filter(r => r.relationshipName === 'department')[0].javadoc = undefined;
```
MathieuAA: I guess this is not useful, can you remove it please?

lib/converters/json_to_jdl_entity_converter.js
pascalgrimaud: done

test/spec/converters/json_to_jdl_entity_converter_test.js
pascalgrimaud: Yes, exactly.
Without the condition in `lib/converters/json_to_jdl_entity_converter.js`, this test will fail, as it can't convert to JDL

MathieuAA:
Thanks :)

PR #278, opened 2018-11-04
Allow relationships from the User entity if skipUserManagement is set

murdos:
If user management is skipped, the User entity can be redefined in jdl, so there's no reason to not allow relationships from it

  - [X] [Travis tests](https://travis-ci.org/jhipster/jhipster-umk/pull_requests) are green


deepu105:
@murdos but I don't think it will work, did you try, I think the generator will not allow you to create an entity with name User even when its skipped. Didn;t check but just a hunch.

murdos:
No, it doesn't work yet, changes are required on the generator side too.

murdos:
Changes on generator-jhipster so the skipUserManagement option is correctly transmitted to jhipster-core.
Currently it is not.

PR #305, opened 2019-02-10
Fix invalid otherEntityRelationshipName generated for the destination association

pvliss:
Hello @MathieuAA. First off thanks for cleaning up `entity_parser.js`.

I have submitted a PR in generator-jhipster to improve the JDL import CI test.: https://github.com/jhipster/generator-jhipster/pull/9217. What I have basically done is merge the previous `app.jdl` with the `big_sample.jdl` used in the core in order to have more test cases at the generator as well.

While testing this today I found that the otherEntityRelationshipName generated for the destination association is invalid hence this PR.

More specifically the following 
```
relationship ManyToMany {
  JobHistory to Job{history},
}
```
generates `otherEntityRelationshipName=jobHistory` for `Job` while it should have been '`otherEntityRelationshipName=job`'

I also changed 
```
relationship ManyToOne {
  Employee{manager(director)} to Employee
}
```
to
```
relationship ManyToOne {
  Employee{manager(lastName)} to Employee
}
``` 
in order to refer to an actual field of the `Employee` entity.

  - [x] [Travis tests](https://travis-ci.org/jhipster/jhipster-umk/pull_requests) are green
  - [x Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/jhipster-core/blob/master/CONTRIBUTING.md) are followed

### DIFF
--- a/lib/parser/entity_parser.js
+++ b/lib/parser/entity_parser.js
@@ -340,7 +340,7 @@ function setDestinationAssociationsForClass(relatedRelationships, entityName) {
     const relationship = {
       relationshipType: _.kebabCase(relationshipType),
       otherEntityName: camelCase(relatedRelationship.from),
-      otherEntityRelationshipName: lowerFirst(otherSplitField.relationshipName) || camelCase(relatedRelationship.from)
+      otherEntityRelationshipName: lowerFirst(otherSplitField.relationshipName) || camelCase(relatedRelationship.to)
     };
     if (relatedRelationship.isInjectedFieldInToRequired) {
       relationship.relationshipValidateRules = REQUIRED;
--- a/test/spec/jdl/jdl_importer_test.js
+++ b/test/spec/jdl/jdl_importer_test.js
@@ -178,7 +178,7 @@ describe('JDLImporter', () => {
               relationshipType: 'many-to-one',
               relationshipName: 'manager',
               otherEntityName: 'employee',
-              otherEntityField: 'director',
+              otherEntityField: 'lastName',
               otherEntityRelationshipName: 'employee'
             },
             {
@@ -246,7 +246,7 @@ describe('JDLImporter', () => {
               relationshipType: 'many-to-many',
               relationshipName: 'history',
               otherEntityName: 'jobHistory',
-              otherEntityRelationshipName: 'jobHistory',
+              otherEntityRelationshipName: 'job',
               otherEntityField: 'id',
               ownerSide: false
             }
--- a/test/spec/parser/entity_parser_test.js
+++ b/test/spec/parser/entity_parser_test.js
@@ -753,7 +753,7 @@ describe('EntityParser', () => {
 
         it('sets one by default', () => {
           expect(content.A.relationships[0].relationshipName).to.equal('b');
-          expect(content.B.relationships[0].otherEntityRelationshipName).to.equal('a');
+          expect(content.B.relationships[0].otherEntityRelationshipName).to.equal('b');
         });
       });
       context('when parsing a relationship with a useJPADerivedIdentifier flag', () => {
--- a/test/test_files/big_sample.jdl
+++ b/test/test_files/big_sample.jdl
@@ -86,7 +86,7 @@ relationship OneToMany {
 
 relationship ManyToOne {
   Employee{user(login)} to User{employee},
-  Employee{manager(director)} to Employee
+  Employee{manager(lastName)} to Employee
 }
 
 relationship ManyToMany {


pvliss:
@MathieuAA I also have one more question if you do not mind. Should the following relationships be unidirectional or bi-directional:

```
relationship ManyToMany {
  JobHistory{department} to Department,
  JobHistory{emp(employee)} to Employee,
}
```

The reason I am asking is that currently jhipster-core treats them as unidirectional and does not generate relationship info for the destination classes but the problem is that at the generator the code generated by `Entity.java.ejs` for `JobHistory` tries to reference these, e.g.:

```java
public class JobHistory implements Serializable {
...
    public JobHistory addDepartment(Department department) {
        this.departments.add(department);
        department.getJobHistories().add(this);
        return this;
    }

    public JobHistory removeDepartment(Department department) {
        this.departments.remove(department);
        department.getJobHistories().remove(this);
        return this;
    }
...
    public JobHistory addEmp(Employee employee) {
        this.emps.add(employee);
        employee.getJobHistories().add(this);
        return this;
    }

    public JobHistory removeEmp(Employee employee) {
        this.emps.remove(employee);
        employee.getJobHistories().remove(this);
        return this;
    }
...
}
```
which in turn leads to a compilation error as these fields are missing from `Department` and `Employee`.

The relevant line at the generator is this: https://github.com/jhipster/generator-jhipster/blob/5ff10fc5b7c156cbddaee253c74bd6b96a19a364/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs#L419

MathieuAA:
Let me check the second case, it could be a big issue this time...

MathieuAA:
This may be a regression. @jdubois quick question: it is possible/wanted to have unidirectional many-to-many relationships, isn't?

pvliss:
@MathieuAA Just to note that I thought that this commit https://github.com/jhipster/generator-jhipster/commit/a72c3c8c5f2688a264e83f968bda11f49012d532 might be causing the problem. So I reverted locally and re-tested but the same problem exists. You can try as well in case I missed something

pvliss:
@MathieuAA Did a bit more investigation and it seems to me that this is not a regression but the generator never supported unidirectional `@ManyToMany` before although they are the default, at least for [JPA](https://en.wikibooks.org/wiki/Java_Persistence/ManyToMany). This is also more apparent by the fact that [all entities used in the CI](https://github.com/jhipster/generator-jhipster/tree/master/test-integration/samples/.jhipster) are bidirectional as well, hence the problem never occurred. I am thinking of adding support for it by adding a new bidirectional attribute to the relationship attribute. This could be done either in jhipster-core or at the generator or both. Would you be interested in such a PR or leave it as is for the moment?

MathieuAA:
That's what I think too but I want to get confirmation from Julien first about this, just to be 100% sure

jdubois:
As far as I know we only support bi-directional association, see #1569 for example.

MathieuAA:
Thanks Julien!

MathieuAA:
And thanks @pvliss :)

pvliss:
Thanks for merging @MathieuAA. 

So are you thinking of getting out a new release and update the jhipster generator as well? Asking as I will also need to update this [jhipster/generator-jhipster#9217](https://github.com/jhipster/generator-jhipster/pull/9217) to remove the `@ManyToMany` unidirectional in there?

Regarding the unidirectional `@ManyToMany`, would you like me to create a PR when I get some time  or are you planning to do it yourself or maybe want to avoid implementing it altogether? I think it would be fairly easy to implement in the generator but of-course it would also add some maintenance overhead as well. In any case I believe that at least the docs should be updated to reflect what is currently supported and avoid confusion.

MathieuAA:
I'm going to do a release tonight Le 13 févr. 2019 19:02, Panayiotis Vlissidis <notifications@github.com> a écrit :Thanks for merging @MathieuAA.
So are you thinking of getting out a new release and update the jhipster generator as well? Asking as I will also need to update this jhipster/generator-jhipster#9217 to remove the @ManyToMany unidirectional in there?
Regarding the unidirectional @ManyToMany, would you like me to create a PR when I get some time  or are you planning to do it yourself or maybe want to avoid implementing it altogether? I think it would be fairly easy to implement in the generator but of-course it would also add some maintenance overhead as well. In any case I believe that at least the docs should be updated to reflect what is currently supported and avoid confusion.

pvliss:
Wow, Just read the issue linked by @jdubois (ashamed that I did not read it before) and it is really interesting although not sure if the performance considerations are still valid. Will look further into it

pvliss:
Hello again @MathieuAA. I see you did a release for [3.6.11](https://www.npmjs.com/package/jhipster-core). Should I make a PR to update the jhipster generator as well?

MathieuAA:
I was planning to tonight, but I don't think I have time. I'll try tomorrow morningLe 15 févr. 2019 20:53, Panayiotis Vlissidis <notifications@github.com> a écrit :Hello again @MathieuAA. I see you did a release for 3.6.11. Should I make a PR to update the jhipster generator as well?

PR #331, opened 2019-05-25
Allow using UUID for all databases, not just Cassandra

murdos:
  - [x] [Travis tests](https://travis-ci.org/jhipster/jhipster-umk/pull_requests) are green

### DIFF
--- a/lib/core/jhipster/field_types.js
+++ b/lib/core/jhipster/field_types.js
@@ -40,6 +40,7 @@ const CommonDBTypes = {
   BIG_DECIMAL: 'BigDecimal',
   FLOAT: 'Float',
   DOUBLE: 'Double',
+  UUID: 'UUID',
   ENUM: 'Enum',
   BOOLEAN: 'Boolean',
   LOCAL_DATE: 'LocalDate',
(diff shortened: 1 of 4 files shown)


MathieuAA:
LGTM, the last PR you need drop is the one for the docs.

murdos:
How can we move forward on the generator-jhipster PR and have integration tests green? It seems we need this PR merged and new JCore release, no?

MathieuAA:
@murdos I don't see why. 
If we merge and release, anyone can write a perfectly good JDL with the new UUID support and get thrown an error when dealing with the generator.
But I see you've written a test using the UUID type... You're doing dependency testing, here. 
Even though I'm generally okay with those, here it's mostly useless as JCore is not some third-party lib :)

pascalgrimaud:
> If we merge and release, anyone can write a perfectly good JDL with the new UUID support and get thrown an error when dealing with the generator.

Not sure about that. As the last generator-jhipster release still uses v3.6.12, which doesn't contain this PR.

murdos:
> 
> 
> @murdos I don't see why.
> If we merge and release, anyone can write a perfectly good JDL with the new UUID support and get thrown an error when dealing with the generator.
> But I see you've written a test using the UUID type... You're doing dependency testing, here.

So I was wrong, I was thinking that jhipster-core was used more intensively in generator-jhipster. So all validation is only done while interpreting JDL to JSON file, but JSON file is not validated when it's loaded. My bad. That's why I'm able to have passing integration tests with JSON. I however note that this would not have been possible with https://github.com/jhipster/generator-jhipster/issues/9651

I however think that the dependency test is useful, since even if both projects are very linked, they have different release cycle. And it's not uncommon to have releases of generator-jhipster that are missing features because the counterpart of the feature in jhipster-core has not been released.
It's also important to have test data in generator-jhipster that are diversified, that's why I added a few UUID.

PR #343, opened 2019-06-30
Implemented the single & multi app export

MathieuAA:
  - [x] [Travis tests](https://travis-ci.org/jhipster/jhipster-core/pull_requests) are green

### DIFF
--- a/lib/converters/json_to_jdl_application_converter.js
+++ b/lib/converters/json_to_jdl_application_converter.js
@@ -23,6 +23,7 @@ const { createJDLApplication } = require('../core/jdl_application_factory');
 const GENERATOR_NAME = 'generator-jhipster';
 
 module.exports = {
+  convertApplicationToJDL,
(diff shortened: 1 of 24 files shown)


MathieuAA:
@Shaolans this is my version of the single & multi app + entities export, can you have a look and compare with yours? if you have the time of course

Shaolans:
The logic behind seems to be similar to me 👍. The major difference is about exporting multi apps, I am exporting on just one jdl as you are exporting a jdl on each apps. I don't know which is the better one, I was thinking that if we have subfoldes containing apps, it sould probably correponds to microservices apps used for an entire project, thus keeping all into one jdl but I might be wrong.

MathieuAA:
Nop, that's a good point! Your logic is better here.

MathieuAA:
@Shaolans it's way better, thanks for the review.

PR #349, opened 2019-07-14
Allows custom annotations for entities

### LINKED ISSUE #320: Allow custom annotations
##### **Overview of the feature request**

Allow custom annotations for metadata/options not handled directly by generator-jhipster.
E.g. : 
```
@myFeature(parameterValue)
entity A
```
or
```
@customAnnotation(myFeature, parameterValue)
entity A
```

It would be nice - and even required if we go to the path of [JDL centric configurations](https://github.com/jhipster/generator-jhipster/issues/8915) - to allow custom annotations for special features added by modules or blueprints.

If you take the example of https://github.com/hipster-labs/generator-jhipster-entity-audit if add a `"enableEntityAudit": true` to the Entity.json file. You currently have to use the jhipster CLI or edit manually Entity.json to enable it.

This can also be very useful for blueprint: by allowing additional specific metadata, the blueprint will be able to make special decision. E.g.: not generate REST API for an entity, ...

##### **Related issues or PR**

- [x] Checking this box is mandatory (this is just to show you read everything)

yelhouti:
Removed checks for unknown options (kept them for built-in ones)
Added related tests
had to change how option `useJPADerivedIdentifier` is used internally, in the JDL it's just: `jpaDerivedIdentifier` which is a PITA. one of these should be deprecated, once json files are removed and we keep only JDL, remove all references of: `useJPADerivedIdentifier`

Fix #320

  - [x] [Travis tests](https://travis-ci.org/jhipster/jhipster-core/pull_requests) are green


MathieuAA:
What happened to the `package-lock.json` file? :)

MathieuAA:
You forgot to export them in the JSON files, for that I suggest having something like that:

```json
{
  "blueprintOptions": {
    "blueprintName": {
      "theAnswer": 42
    }
}
```

yelhouti:
> What happened to the `package-lock.json` file? :)

nothing on purpose XD

> You forgot to export them in the JSON files, for that I suggest having something like that:

@MathieuAA Indeed I when I tried to use it with the generator this afternoon, I couldn't they don't get exported. working on it. I wouldn't even tell the blueprint, they can use whatever they want, I'll just add it like this:
```
{
    "name": "Customer",
    "fields": [
        {
            "fieldName": "name",
            "fieldType": "String",
            "fieldValidateRules": [
                "required"
            ]
        },
        id: true
    ],
    "relationships": [
        {
            "relationshipType": "one-to-many",
            "otherEntityName": "task",
            "otherEntityRelationshipName": "customer",
            "relationshipName": "task"
        },
        id: true
    ],
...
}
....
```
for a jdl like this:
```
entity Customer {
    @id
    name String required
}

relationship OneToMany {
    Customer{@id task} to Task{customer(name) required}
}
...
```

yelhouti:
@MathieuAA annotations today are suppose to be before the javadoc, do you mind me moving them after this is a BREAKING CHANGE

yelhouti:
@MathieuAA I found this annoying thing where: the jdl uses `jpaDerivedIdentifier` and the json uses `useJpaDerivedIdentifier` since the new code is generic (and uses with and annotation the same way) which one do you suggest I keep?

yelhouti:
@MathieuAA you can review and merge if you have time, thanks a lot :)

deepu105:
Please don't introduce breaking changes. As we can't make a major release
for generator

yelhouti:
thanks for your ansewer @deepu105 , glad I didn't, maybe we can keep switching comments and annotations for another PR, if you (all of you) are ok with it.

lib/parsers/entity_parser.js
```diff
@@ -183,7 +183,7 @@ function setEntityNamesOptions(option) {
         );
         break;
       default:
-        entities[entityName][option.name] = option.value;
+        entities[entityName][option.name] = option.value ? option.value : true;
```
MathieuAA: `entities[entityName][option.name] = option.value || true;`

lib/parsers/entity_parser.js
```diff
@@ -274,6 +275,15 @@ function setValidationsOfField(jdlField, fieldData) {
   }
 }
 
+function setOptionsOfField(jdlField, fieldData) {
+  if (Object.entries(jdlField.options).length !== 0) {
+    fieldData.options = {};
+    Object.entries(jdlField.options).forEach(option => {
```
MathieuAA: What you can do here:
```javascript
Object.entries(jdlField.options).forEach(([key, value]) => ...)
```

lib/parsers/entity_parser.js
```diff
@@ -376,10 +384,26 @@ function setDestinationAssociationsForClass(relatedRelationships, entityName) {
...
       relationship.ownerSide = false;
     }
+    setOptionsOfRelationship(relatedRelationship, relationship);
     entities[entityName].addRelationship(relationship);
   });
 }
 
+function setOptionsOfRelationship(relatedRelationship, relationship) {
+  if (relatedRelationship.options) {
+    Object.entries(relatedRelationship.options).forEach(option => {
```
MathieuAA: same remark, you can do this `Object.entries(jdlField.options).forEach(([key, value]) => ...)`

lib/converters/json_to_jdl_entity_converter.js
```diff
@@ -189,7 +189,7 @@ function getRelationship(relationship, entityName) {
     commentInTo: null,
     otherEntity: null,
     toJdlEntity: null,
-    options: new Set()
+    options: []
```
MathieuAA: why?

lib/core/jdl_entity.js
```diff
@@ -32,6 +32,7 @@ class JDLEntity {
     this.tableName = merged.tableName || merged.name;
     this.fields = merged.fields;
     this.comment = merged.comment;
+    this.options = merged.options;
```
MathieuAA: What are options exactly? Strings or objects? If they are strings, why not have a Set?

lib/core/jdl_field.js
```diff
@@ -52,6 +52,7 @@ class JDLField {
      */
     Object.values(merged.validations).forEach(checkValidation);
     this.validations = merged.validations;
+    this.options = merged.options;
```
MathieuAA: Options in the JDL entity class is an array, here an object.
Second issue, I like to use objects when the target isn't a collection (if it is, I use a Set or a Map)

lib/parsers/document_parser.js
```diff
@@ -305,6 +303,26 @@ function fillAssociations() {
   }
 }
 
+function convertAnnotationsToOptions(annotations) {
+  const result = {};
+  annotations.forEach(o => {
```
MathieuAA: could you not use `o` as variable name here please?

lib/parsers/document_parser.js
```diff
@@ -305,6 +303,26 @@ function fillAssociations() {
...
+    const value = o.method ? o.method : true;
```
MathieuAA: you can use here `o.method || true`
by the way, why true?

lib/parsers/document_parser.js
```diff
@@ -305,6 +303,26 @@ function fillAssociations() {
...
+    if (o.option in result) {
+      const oldValue = result[o.option];
+      if (Array.isArray(oldValue)) {
+        if (value !== oldValue) {
```
MathieuAA: I'm not sure it works.
Here, if I'm getting this right, `oldValue` is an array, right? so comparing an array if another won't work (try [1] === [1] for instance)

lib/parsers/entity_parser.js
```diff
@@ -274,6 +275,15 @@ function setValidationsOfField(jdlField, fieldData) {
   }
 }
 
+function setOptionsOfField(jdlField, fieldData) {
```
MathieuAA: I would have used ...forField

lib/parsers/entity_parser.js
```diff
@@ -340,6 +347,7 @@ function setSourceAssociationsForClass(relatedRelationships, entityName) {
 
       relationship.ownerSide = true;
     }
+    setOptionsOfRelationship(relatedRelationship, relationship);
```
MathieuAA: For here instead of "of", what do you think?

.nycrc.json
```diff
@@ -4,8 +4,6 @@
   "exclude": [
     "lib/dsl/gen/*.js",
     "lib/dsl/self_checks/*.js",
-    "lib/dsl/ast_builder.js",
```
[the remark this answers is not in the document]
yelhouti: Well I cam remove this, I just don't understand why it was ignored from the test report whle it is well tested

lib/parsers/entity_parser.js
yelhouti: nice catch

lib/parsers/entity_parser.js
yelhouti: indeed destructing is much better, doing this

lib/parsers/document_parser.js
yelhouti: sure

lib/parsers/document_parser.js
yelhouti: well annotations have a key and a value, some annotations do not need a value ex: `@id`, you would just check `if (annotation.method)` which would evaluate to `true`.

lib/parsers/document_parser.js
yelhouti: Je sais bien, je pense on relisant que je cherchais à faire ça:
`if (!oldValue.includes(value)) {` j'étais décidément très fatigué, merci d'avoir vu.

lib/parsers/entity_parser.js
yelhouti: En effet

lib/parsers/entity_parser.js
yelhouti: Sure

lib/core/jdl_field.js
yelhouti: Indeed, for classes it has always been array and i can't change this, without breaking everything that depends on it (this is a public API so many blueprints modules... would stop working if I change this), I also think it an array because in the beginning you only had Unary options, since we use options with keys and values an object makes more sense

lib/converters/json_to_jdl_entity_converter.js
yelhouti: because `Set` works with shallow equals, therefore adding the same option twicefor non UnaryOptions will pass the test. You could ask why not a Map/Object because the same annotation can be used twice. For places where this was the case, it's still is. for places where I added options, I preferred keeping it simple and directly expose the object.

lib/core/jdl_entity.js
yelhouti: It depends on where you are checking, and I totally agree it's a bit messed up, but if you check `jdl_options.js` (and I didn't touch that file), options are stored in an object (used as a Map), but when getOptions is called and array is returned. 
For places where this was the case (options did exist), it's still the sam (to avoid breaking changes in the public API). for places where I added options, I preferred keeping it simple and directly exposing the object.

lib/parsers/entity_parser.js
yelhouti: done

.nycrc.json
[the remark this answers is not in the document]
yelhouti: I ll let you decide, if it was for me, i would leave it, since it's our code...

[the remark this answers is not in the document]
MathieuAA: There could be another test entirely, but the value brought by this PR is too high just for this (and I could see a test for this file)
I'll create an issue to create a test file for this

yelhouti:
IMO this file should not even be versioned, it changes on each compilation, but I can recompile whenever you are ready

MathieuAA:
It changes if the grammar changes, and is needed for web projects that depend on JCore (the JDL studio). If you think of a better workflow, I'd be more than happy to hear it!!

.nycrc.json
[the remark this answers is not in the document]
yelhouti: removing

yelhouti:
other porject having jdl-core as a dependency should be enough no?

MathieuAA:
How so? I don't get your question

yelhouti:
> How so? I don't get your question

gitter?

MathieuAA:
yep

murdos:
Awesome!!! 👍 Thanks @yelhouti

clement26695:
This seems really cool ! Would it be possible to update the docs to explain what can be done, and how to use this ?

yelhouti:
@murdos my pleasure
@clement26695 as soon as it is released, this mainly useful for blueprints, in my use case I used it for composite ids:
https://www.npmjs.com/package/generator-jhipster-composite-key-server
https://www.npmjs.com/package/generator-jhipster-primeng-blueprint

PR #373, opened 2019-09-22
Fixed annotation/comment entity/field flow

### LINKED ISSUE #369: Order between annotation and javadoc is significant and can lead to missing javadoc or crash
##### **Overview of the issue**

While doing some tests with the customs annotations from https://github.com/jhipster/jhipster-core/pull/349 from master version, I noticed several issues related to the order of javadoc and annotations:

**On entity level:**

Works:
```
@dto(mapstruct)
/** entity javadoc */
entity A {
  /** field javadoc */ name String
}
```

However with the following jdl, the javadoc is silently dropped:
```
/** entity javadoc */
@dto(mapstruct)
entity A {
  /** field javadoc */ name String
}
```

**On field level:**

Works:
```
/** entity javadoc */
entity A {
  @id /** field javadoc */ name String
}
```

Crashes:
```
/** entity javadoc */
entity A {
  /** field javadoc */ 
  @id name String
}
```
with the error:
```
INFO! The JDL is being parsed.
error: Error message:
	MismatchedTokenException: Expecting token of type --> NAME <-- but found --> '@' <--
	at line: 4, column: 3
SyntaxError: Error: MismatchedTokenException: Expecting token of type --> NAME <-- but found --> '@' <--
	at line: 4, column: 3
Error while parsing applications and entities from the JDL SyntaxError: Error: MismatchedTokenException: Expecting token of type --> NAME <-- but found --> '@' <--
	at line: 4, column: 3
SyntaxError: Error: MismatchedTokenException: Expecting token of type --> NAME <-- but found --> '@' <--
```

Similarly to what’s possible in Java, it should be possible to add javadoc before annotation and  the other way round.
The javadoc should not be silently dropped or makes the parser crashes.

##### **Reproduce the error**

See above

##### **Suggest a Fix**

No idea! 😁

##### **JHipster Version(s)**

6.3.0 with jhipster-core master

##### **Browsers and Operating System**

- [X] Checking this box is mandatory (this is just to show you read everything)

MathieuAA:
Before, one could annotate then comment entities and fields and sometimes it crashed sometimes it didn't parse the comment.
Now, only commenting then annotating works and is supported.
This is a breaking change.

Also introduced grammar tests.

Fixes #369

  - [ ] Checks are green


test/spec/grammar/grammar.js
```diff
@@ -0,0 +1,139 @@
+const { expect } = require('chai');
+const { parseFromContent } = require('../../../lib/readers/jdl_reader');
+
+describe('Grammar tests', () => {
```
murdos: I noticed most (but not all) tests are in files named `*_test.js`. 
It seems `spec.js` suffix is a bit more standard. Would you accept a PR renaming test files to use `spec.js` suffix?

MathieuAA: yes of course!

PR #402, opened 2019-11-17
Added more cases to the relationship validator

MathieuAA:
Cases:
  - MtO:
    - if the User entity has the injected field without the skipUserMgt option set, it should fail
  - MtM:
    - if there is an unidirectional relationship to the User, it should not fail as this is a valid case

@pascalgrimaud I've shamelessly taken the sample from the generator for my tests, ideally this project should have been included in the generator and this mess could have been avoided... :(

### DIFF
--- a/lib/exceptions/business_error_checker.js
+++ b/lib/exceptions/business_error_checker.js
@@ -148,17 +148,14 @@ class BusinessErrorChecker {
   }
 
   checkForRelationshipErrors() {
-    const validator = new RelationshipValidator();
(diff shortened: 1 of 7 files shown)


pascalgrimaud:
no worry
it's because the daily builds doesn't cover all cases. I need to find time to put there better JDL

MathieuAA:
@pascalgrimaud I've got an idea: I'm gonna use a merged project (generator + JCore) locally and modify the "blob" (I don't have a name for the merge of these two) to test everything that I need.

PR #409, opened 2019-12-09
Implement creationTimestamp for reproducible builds.

mshima:
  - [ ] Checks are green


mshima:
This patch will need generator-jhipster to pass creationTimestamp

lib/parsers/entity_parser.js
```diff
@@ -107,7 +107,16 @@ function resetState() {
   isType = null;
 }
 
-function initializeEntities() {
+function initializeEntities(args) {
+  let creationTimestamp;
```
MathieuAA: might as well initialize this variable with `new Date()`

lib/parsers/entity_parser.js
```diff
@@ -107,7 +107,16 @@ function resetState() {
...
+  if (args.creationTimestamp) {
```
MathieuAA: here we're not really initializing entities: we're setting up the creation timestamp
it should belong to another function

lib/utils/format_utils.js
```diff
@@ -62,7 +62,8 @@ function dateFormatForLiquibase(args) {
     args.date = new Date(args.date.getTime());
   }
   const merged = merge(defaultsForLiquibaseDateFormatting(), args);
-  merged.date.setSeconds(merged.date.getUTCSeconds() + merged.increment);
+  merged.date.setMinutes(merged.date.getMinutes() + merged.increment);
```
MathieuAA: why?

mshima: Used minutes at generator-jhipster patch. 
https://github.com/jhipster/generator-jhipster/blob/1db47508bb9a89d16ef5798bee0ee9c63c9300a1/generators/generator-base.js#L822
This will keep a delta for custom changelogs.

Removed UTC because getUTC is already used next.
https://github.com/jhipster/jhipster-core/blob/d3193b5d5c7d2e66eb75fe9ef34f3fb5e8d106b5/lib/utils/format_utils.js#L72

Running getUTCSeconds from a date with seconds set with a getUTCSeconds will duplicate conversion if it exists.

MathieuAA: fair enough

lib/exceptions/application_validator.js
```diff
@@ -87,6 +87,9 @@ function checkSpecificOptions(option, value) {
     case 'prodDatabaseType':
       checkProdDatabaseTypeValue(value);
       break;
+    case 'creationTimestamp':
```
MathieuAA: that implies that the creationTimestamp is an app option, I don't think it is, is it?

lib/jdl/jdl_importer.js
```diff
@@ -76,6 +76,16 @@ function makeJDLImporter(content, configuration) {
     exportedDeployments: []
   };
 
+  configuration.creationTimestampConfig = new Date();
```
MathieuAA: I like it better here instead of the entity_parser

lib/jdl/jdl_importer.js
```diff
@@ -102,6 +103,18 @@ function makeJDLImporter(content, configuration) {
   };
 }
 
+function parseConfiguration(configuration) {
```
MathieuAA: it doesn't exactly do what its name implies, I'd have named it "setupCreationTimestampFromConfiguration"

mshima: What about "setupConfiguration" to keep it generic.

lib/exceptions/application_validator.js
mshima: What do you mean about app option? jdl option?

mshima: generator-jhipster should forward it if passed as an arg.

mshima: This is not needed.

lib/jdl/jdl_importer.js
MathieuAA: Yes, let's

lib/exceptions/application_validator.js
MathieuAA: in the .yo-rc file too?

mshima: It writes to .yo-rc file for a rebuild with `jhipster --with-entities`

MathieuAA: okay then, it's an acceptable coupling for me

MathieuAA:
@mshima If I'm not mistaken, this PR can be merged right?

mshima:
Yes

PR #418, opened 2019-12-19
Do not allow underscore in baseName to prevent invalid hostnames

### LINKED ISSUE #417: Underscores in UAA base name are allowed causing problems when running
##### **Overview of the issue**

I'm able to generate a project from a .jh file with a `baseName` like `xxx_uaa` for the UAA microservice (or any other microservice). However, the microservices will run with an invalid hostname causing `IllegalStateException: Request URI does not contain a valid hostname` when contacting it.

Well, it hurts to replace all microservices' names after generation. Also, if underscores can't be used for the reason mentioned above, maybe there should exist another way to prefix basenames? Hyphens would be a good candidate (not sure why they were suppressed in the past though).

##### **Reproduce the error**

1. Create a microservice with an underscore in the name
2. Create another microservice that uses @FeignClient("<service name with underscore>")
3. Make a call to the FeignClient

##### **Related issues**

https://github.com/jhipster/generator-jhipster/issues/3984

##### **Suggest a Fix**

- Add error message when user tries to generate microservice with an underscore or illegal character according to URI spec in the service name.
- Or... replace underscore with hyphen for hostnames
- Or... 1st option + allow hyphens instead

##### **JHipster Version(s)**

JHipster v6.5.1

##### **JHipster configuration**

openjdk version "11.0.5" 2019-10-15
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.5+10)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.5+10, mixed mode)

git version 2.20.1

node: v12.13.1

npm: 6.13.4

yeoman: 3.1.1

Docker version 19.03.3, build a872fc2f86

docker-compose version 1.24.1, build 4667896b

-   [x] Checking this box is mandatory (this is just to show you read everything)

josepaiva94:
  - [x] Checks are green

Fixes #417 with the suggested fix "Add error message when user tries to generate microservice with an underscore or illegal character according to URI spec in the service name."

However, i would like to use dashes. May I make another PR or is there a good reason against dashes in baseName?

### DIFF
--- a/lib/dsl/validator.js
+++ b/lib/dsl/validator.js
@@ -59,7 +59,7 @@ const configPropsValidations = {
   },
   BASE_NAME: {
     type: 'NAME',
-    pattern: ALPHANUMERIC_UNDERSCORE,
+    pattern: ALPHANUMERIC,
     msg: 'baseName property'
   },
   BLUEPRINT: {
--- a/test/spec/jdl/jdl_importer.spec.js
+++ b/test/spec/jdl/jdl_importer.spec.js
@@ -1119,24 +1119,17 @@ describe('JDLImporter', () => {
       });
     });
     context('when parsing a JDL with underscores contained in the application name', () => {
-      let returned;
-
-      before(() => {
-        const importer = createImporterFromFiles([path.join('test', 'test_files', 'underscore_application_name.jdl')]);
-        returned = importer.import();
-      });
-
       after(() => {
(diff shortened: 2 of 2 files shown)


MathieuAA:
Hello, your fix disallow underscores for every type of app, not just MS.
Yes, another PR for dashes if they are allowed.

PR #439, opened 2020-03-13
Add returnMergedEntity option

mshima:
  - [ ] Checks are green

### DIFF
--- a/lib/exporters/jhipster_entity_exporter.js
+++ b/lib/exporters/jhipster_entity_exporter.js
@@ -89,17 +89,17 @@ function getEntitiesToExport(entityNamesToExport, entities) {
  */
 function exportEntities(passedConfiguration) {
   init(passedConfiguration);
-  createJHipsterJSONFolder(
-    passedConfiguration.application.forSeveralApplications ? configuration.application.name : ''
-  );
+  const subFolder = passedConfiguration.application.forSeveralApplications ? configuration.application.name : '';
+  createJHipsterJSONFolder(subFolder);
+  configuration.entities = updateEntities(subFolder);
   if (!configuration.forceNoFiltering) {
-    configuration.entities = filterOutUnchangedEntities();
+    configuration.entities = filterOutUnchangedEntities(subFolder);
   }
   if (shouldFilterOutEntitiesBasedOnMicroservice()) {
     configuration.entities = filterOutEntitiesByMicroservice();
   }
   if (!passedConfiguration.skipEntityFilesGeneration) {
-    writeEntities(passedConfiguration.application.forSeveralApplications ? configuration.application.name : '');
+    writeEntities(subFolder);
   }
   return configuration.entities;
 }
@@ -123,11 +123,21 @@ function createJHipsterJSONFolder(subFolder) {
  * Writes entities in a sub folder.
  * @param subFolder the folder (to create) in which the JHipster entity folder will be.
  */
+function updateEntities(subFolder) {
+  return configuration.entities.map(entity => {
+    const filePath = path.join(subFolder, toFilePath(entity.name));
+    return updateEntityToGenerateWithExistingOne(filePath, entity);
+  });
+}
+
+/**
+ * Writes entities in a sub folder.
+ * @param subFolder the folder (to create) in which the JHipster entity folder will be.
+ */
 function writeEntities(subFolder) {
   configuration.entities.forEach(entity => {
     const filePath = path.join(subFolder, toFilePath(entity.name));
-    const updatedEntity = updateEntityToGenerateWithExistingOne(filePath, entity);
-    fs.writeFileSync(filePath, JSON.stringify(updatedEntity, null, 2).concat('\n'));
+    fs.writeFileSync(filePath, JSON.stringify(entity, null, 2).concat('\n'));
   });
 }
 
@@ -142,13 +152,10 @@ function updateEntityToGenerateWithExistingOne(filePath, entity) {
   return entity;
 }
 
-function filterOutUnchangedEntities() {
+function filterOutUnchangedEntities(subFolder) {
   return configuration.entities.filter(entity => {
-    const filePath = toFilePath(entity.name);
-    return !(
-      FileUtils.doesFileExist(filePath) &&
-      areJHipsterEntitiesEqual(readJSONFile(filePath), updateEntityToGenerateWithExistingOne(filePath, entity))
-    );
+    const filePath = path.join(subFolder, toFilePath(entity.name));
+    return !(FileUtils.doesFileExist(filePath) && areJHipsterEntitiesEqual(readJSONFile(filePath), entity));
   });
 }
 
--- a/test/spec/exporters/jhipster_entity_exporter.spec.js
+++ b/test/spec/exporters/jhipster_entity_exporter.spec.js
@@ -399,6 +399,7 @@ describe('JHipsterEntityExporter', () => {
         context('inside a microservice', () => {
           context('and when entities without the microservice option are passed', () => {
             let entities;
+            let returnedContent;
 
             before(() => {
               entities = [
@@ -501,7 +502,7 @@ describe('JHipsterEntityExporter', () => {
                   name: 'G'
                 }
               ];
-              JHipsterEntityExporter.exportEntities({
+              returnedContent = JHipsterEntityExporter.exportEntities({
                 entities,
                 application: {
                   name: 'client',
@@ -516,6 +517,13 @@ describe('JHipsterEntityExporter', () => {
               });
             });
 
+            it('returns every entity', () => {
+              expect(returnedContent.length).to.be.equal(7);
+              ['A', 'B', 'C', 'D', 'E', 'F', 'G'].forEach(entityName => {
+                expect(returnedContent.filter(entity => entity.name === entityName) !== undefined).to.be.true;
+              });
+            });
+
             after(() => {
               ['A', 'B', 'C', 'D', 'E', 'F', 'G'].forEach(entityName => {
                 fs.unlinkSync(`.jhipster/${entityName}.json`);
@@ -525,6 +533,7 @@ describe('JHipsterEntityExporter', () => {
           });
           context('and when microservice entities are passed', () => {
             let entities;
+            let returnedContent;
 
             before(() => {
               entities = [
@@ -589,7 +598,7 @@ describe('JHipsterEntityExporter', () => {
                   name: 'Product'
                 }
               ];
-              JHipsterEntityExporter.exportEntities({
+              returnedContent = JHipsterEntityExporter.exportEntities({
                 entities,
                 application: {
                   name: 'client',
@@ -603,6 +612,10 @@ describe('JHipsterEntityExporter', () => {
               expect(FileUtils.doesFileExist('.jhipster/Location.json'));
             });
 
+            it('only returns the entities that should be inside the microservice', () => {
+              expect(returnedContent.length).to.be.equal(2);
+            });
+
             after(() => {
               fs.unlinkSync('.jhipster/Client.json');
               fs.unlinkSync('.jhipster/Location.json');
@@ -614,6 +627,7 @@ describe('JHipsterEntityExporter', () => {
       context('when exporting updated entities', () => {
         let originalContent;
         let newContent;
+        let returnedContent;
 
         before(() => {
           originalContent = {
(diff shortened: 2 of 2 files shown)


lib/exporters/jhipster_deployment_exporter.js
```diff
@@ -49,7 +49,7 @@ function checkForErrors(deployment) {
 }
 
 function setUpDeploymentStructure(deployment) {
-  const cleanDeployment = JSON.parse(JSON.stringify(deployment, null, 2));
+  const cleanDeployment = JSON.parse(JSON.stringify(deployment, null, 2).concat('\n'));
```
MathieuAA: The `cleanDeployment` name wasn't great... let's just remove the variable which is only used once and do l.54:
`deploymentToExport[GENERATOR_NAME] = JSON.parse(JSON.stringify(deployment, null, 2).concat('\n'));`

lib/exporters/jhipster_application_exporter.js
```diff
@@ -49,10 +49,12 @@ function exportApplications(applications) {
  * @param application, the JDL application to exporters.
  * @return the exported application in its final form.
  */
-function exportApplication(application) {
+function exportApplication(application, configuration = {}) {
```
MathieuAA: I like it, passing it in a configuration object is a good idea!

lib/exporters/jhipster_entity_exporter.js
```diff
@@ -92,7 +93,9 @@ function exportEntities(passedConfiguration) {
   if (shouldFilterOutEntitiesBasedOnMicroservice()) {
     filterOutEntitiesByMicroservice();
   }
-  writeEntities(passedConfiguration.application.forSeveralApplications ? configuration.application.name : '');
+  if (!passedConfiguration.configuration || !passedConfiguration.configuration.skipEntityFilesGeneration) {
```
MathieuAA: Might as well create a variable for that `!passedConfiguration.configuration || !passedConfiguration.configuration.skipEntityFilesGeneration`, `skippedEntityFileGenerations`

MathieuAA: Can you update the doc for this attribute `skipEntityFilesGeneration` please?

lib/exporters/jhipster_application_exporter.js
```diff
@@ -49,10 +49,12 @@ function exportApplications(applications) {
...
   checkForErrors(application);
   const exportableApplication = setUpApplicationStructure(application);
-  writeConfigFile(exportableApplication);
+  if (!configuration.skipYoRcGeneration) {
```
MathieuAA: can you add the `configuration` & `skipYoRcGeneration` attributes to the doc please?

mshima: I was testing back there, forgot to add when I create the PR.
Added to JSDoc, but used a new format, can use a old format if you want.

MathieuAA: Nop that's good!

lib/exporters/jhipster_application_exporter.js
```diff
@@ -46,13 +46,17 @@ function exportApplications(applications) {
 
 /**
  * Exports JDL a application to a JDL file in the current directory.
- * @param application, the JDL application to exporters.
+ * @param {Object}  application, the JDL application to exporters.
```
MathieuAA: typo: to export

lib/exporters/jhipster_application_exporter.js
```diff
@@ -46,13 +46,17 @@ function exportApplications(applications) {
...
+ * @param {Boolean} [configuration.skipYoRcGeneration], set true to skip writing .yo-rc.json.
  * @return the exported application in its final form.
  */
-function exportApplication(application) {
+function exportApplication(application, configuration = {}) {
   checkForErrors(application);
   const exportableApplication = setUpApplicationStructure(application);
-  writeConfigFile(exportableApplication);
+  if (!configuration.skipYoRcGeneration) {
+    writeConfigFile(exportableApplication);
```
MathieuAA: this needs testing too

lib/exporters/export_utils.js
```diff
@@ -36,14 +36,19 @@ function writeConfigFile(config, yoRcPath = '.yo-rc.json') {
...
         ...yoRc[GENERATOR_NAME],
-        ...config[GENERATOR_NAME]
+        ...config[GENERATOR_NAME],
+        creationTimestamp
       }
     };
   }
-  fs.writeFileSync(yoRcPath, JSON.stringify(newYoRc, null, 2));
+  fs.writeFileSync(yoRcPath, JSON.stringify(newYoRc, null, 2).concat('\n'));
 }
```
MathieuAA: this needs testing too

lib/jdl/jdl_importer.js
```diff
@@ -184,7 +184,7 @@ function importOneApplicationAndEntities(jdlObject, configuration) {
     exportedDeployments: []
   };
   const jdlApplication = jdlObject.applications[Object.keys(jdlObject.applications)[0]];
-  importState.exportedApplications.push(JHipsterApplicationExporter.exportApplication(jdlApplication));
+  importState.exportedApplications.push(JHipsterApplicationExporter.exportApplication(jdlApplication, configuration));
```
MathieuAA: this needs testing too

mshima: I couldn't think test for importOneApplicationAndEntities, exportApplication is duplicated

MathieuAA: oh okay! I'll see about that, thanks

lib/exporters/jhipster_application_exporter.js
mshima: Done

lib/exporters/export_utils.js
mshima: Done

test/spec/exporters/jhipster_entity_exporter.spec.js
```diff
@@ -767,7 +767,7 @@ describe('JHipsterEntityExporter', () => {
           fs.rmdirSync('.jhipster');
         });
 
-        it('creates the entity in the current folder', () => {
+        it("don't creates the entity in the current folder", () => {
```
MathieuAA: doesn't create

test/spec/exporters/jhipster_application_exporter.spec.js
```diff
@@ -181,7 +205,7 @@ describe('JHipsterApplicationExporter', () => {
           fs.unlinkSync(path.join('.yo-rc.json'));
         });
 
-        it('adds the read content to the exported application', () => {
+        it('adds the read content to the exported application and keep selected', () => {
```
MathieuAA: 2 things:
  - what does keep selected mean?
  - if there's an "and" in an "it", it usually means you're testing two things and this should be split in two "it"

test/spec/exporters/jhipster_application_exporter.spec.js
```diff
@@ -148,6 +148,28 @@ describe('JHipsterApplicationExporter', () => {
...
+              enableTranslation: false,
+              languages: ['en', 'fr'],
+              jhipsterVersion: '4.9.0',
+              otherModules: ['MyModule']
+            }),
+            { skipYoRcGeneration: true }
+          );
+        });
+
+        it("don't creates the .yo-rc.json in the current folder", () => {
```
MathieuAA: doesn't create*

test/spec/exporters/jhipster_application_exporter.spec.js
```diff
@@ -175,13 +199,7 @@ describe('JHipsterApplicationExporter', () => {
...
             })
           );
-          content = JSON.parse(fs.readFileSync(path.join('.yo-rc.json'), { encoding: 'utf8' }));
-        });
-        after(() => {
-          fs.unlinkSync(path.join('.yo-rc.json'));
-        });
-
-        it('adds the read content to the exported application', () => {
+          const content = JSON.parse(fs.readFileSync(path.join('.yo-rc.json'), { encoding: 'utf8' }));
```
MathieuAA: I don't like having logic inside an it, "it"s should only have "expect"s, and setup code inside a before/beforeEach

test/spec/exporters/jhipster_application_exporter.spec.js
```diff
@@ -221,6 +239,35 @@ describe('JHipsterApplicationExporter', () => {
...
+              applicationType: MONOLITH,
+              baseName: 'toto',
+              packageName: 'com.mathieu.sample',
+              enableTranslation: false,
+              languages: ['en', 'fr'],
+              jhipsterVersion: '4.9.0',
+              creationTimestamp: 'new'
+            })
+          );
+          const content = JSON.parse(fs.readFileSync(path.join('.yo-rc.json'), { encoding: 'utf8' }));
```
MathieuAA: this should be inside a before

MathieuAA:
Can you finish this PR please @mshima?

mshima:
@MathieuAA sure. I will try later today.

mshima:
Probably I will split in smaller PRs.

MathieuAA:
Thanks :) Do as you please!

mshima:
@MathieuAA first round is ready.

MathieuAA:
Can we close this one then?

mshima:
> Can we close this one then?

There are 2 more changes.

mshima:
@MathieuAA this is the last one for now.

.gitignore
```diff
@@ -7,9 +7,9 @@ out/
...
 .sonar
-coverage/*
-.nyc_output/*
+coverage/
+.nyc_output/
 yarn-error.log
 jhipster-core.iml
 .project
-.vscode/*
+.vscode/
```
MathieuAA: Not part of the PR, you can directly merge to master for this, it's easier

lib/exporters/jhipster_entity_exporter.js
```diff
@@ -124,7 +123,13 @@ function writeEntities(subFolder) {
   for (let i = 0, entityNames = Object.keys(configuration.entities); i < entityNames.length; i++) {
     const filePath = path.join(subFolder, toFilePath(entityNames[i]));
     const entity = updateEntityToGenerateWithExistingOne(filePath, configuration.entities[entityNames[i]]);
-    fs.writeFileSync(filePath, JSON.stringify(entity, null, 2).concat('\n'));
+    if (configuration.returnMergedEntity) {
```
MathieuAA: Shouldn't it always be true? I mean, I don't see cases where we don't want that...

lib/exporters/jhipster_entity_exporter.js
```diff
@@ -124,7 +123,13 @@ function writeEntities(subFolder) {
...
+      // Set the updated entity back.
+      configuration.entities[entityNames[i]] = entity;
+    }
+    if (!configuration.skipEntityFilesGeneration) {
```
MathieuAA: Having a conditional that says "if we don't want to write entity files" in a function that's called `writeEntities` is weird, don't you think?
Having two functions makes more sense.

lib/exporters/jhipster_entity_exporter.js
mshima: Your call.
I did for backward compatibility.
But since this can be considered a bug, a minor release should be enough for this change.
The only use case I can think is if some blueprint wants to change the behavior. But I don’t think we should support that.

lib/exporters/jhipster_entity_exporter.js
mshima: I agree but if we follow the rules, this change should be a major release.

lib/exporters/jhipster_entity_exporter.js
MathieuAA: Ah! good point. You're absolutely right! Let's make it the default, we'll do release including this later (I still have to finish the "options in app" thing)

lib/exporters/jhipster_entity_exporter.js
MathieuAA: No need, it's a private function. Had it been in the exports then yes.
How do you see it? Maybe we're not thinking about the same things :)

My idea is : `exportEntities` should call two functions. The first one being the "we want to write files to the disk" and the other "the flag is set, let's not write"

mshima: I am thinking about the following approaches:
- Create a new mergeExistingEntityFromFile.
Duplicate current writeEntities removing the write part.
- Rename the function to mergeAndWriteEntities and don’t duplicate the loop

mshima:
I just noticed that the only method that take subFolder (application.name) into account is writeEntities.
Am I wrong or should we change that?

MathieuAA:
> I just noticed that the only method that take subFolder (application.name) into account is writeEntities.
Am I wrong or should we change that?

@mshima sorry didn't see your comment... Anyway, this is good to go, thanks :)

PR #443, opened 2020-03-26
Keep original creationTimestamp

mshima:
  - [ ] Checks are green

### DIFF
--- a/lib/exporters/export_utils.js
+++ b/lib/exporters/export_utils.js
@@ -36,12 +36,17 @@ function writeConfigFile(config, yoRcPath = '.yo-rc.json') {
   let newYoRc = { ...config };
   if (FileUtils.doesFileExist(yoRcPath)) {
     const yoRc = JSON.parse(fs.readFileSync(yoRcPath, { encoding: 'utf-8' }));
+    let creationTimestamp = config[GENERATOR_NAME].creationTimestamp;
+    if (yoRc[GENERATOR_NAME] && yoRc[GENERATOR_NAME].creationTimestamp) {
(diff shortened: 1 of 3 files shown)


lib/exporters/export_utils.js
```diff
@@ -36,12 +36,17 @@ function writeConfigFile(config, yoRcPath = '.yo-rc.json') {
...
+      creationTimestamp = yoRc[GENERATOR_NAME].creationTimestamp;
+    }
     newYoRc = {
       ...yoRc,
       ...config,
       [GENERATOR_NAME]: {
         ...yoRc[GENERATOR_NAME],
-        ...config[GENERATOR_NAME]
+        ...config[GENERATOR_NAME],
+        creationTimestamp
```
MathieuAA: can you add the logic in the tests corresponding to this file please

mshima: I don’t get.

mshima: Do you mean split into a separated test?

MathieuAA: This snippet does this:
  - if there's a yo-rc file, and the creationTimestamp is present, it uses it (case 1)
  - otherwise, it uses the one from the config (case 2)
  - if both are absent, it does not set it (case 3)

These are the cases I need mentioned in tests for export_utils

MathieuAA: You wrote the tests for the exporter (likely an integration test), I want unit tests too

mshima: Where they are implemented?

MathieuAA: The tests? Under the spec folder. Search for export_utils.spec

PR #446, opened 2020-04-01
Skip Hard Stop Error Message When OptionType Does Not Exist

SudharakaP:
When some option type is missing we do not hard stop. Instead we log that the option type is missing and move on.

Fixes https://github.com/jhipster/generator-jhipster/issues/11532

  - [ ] Checks are green

### DIFF
--- a/lib/core/jdl_application_configuration_factory.js
+++ b/lib/core/jdl_application_configuration_factory.js
@@ -22,7 +22,14 @@ const StringJDLApplicationConfigurationOption = require('./string_jdl_applicatio
 const IntegerJDLApplicationConfigurationOption = require('./integer_jdl_application_configuration_option');
 const BooleanJDLApplicationConfigurationOption = require('./boolean_jdl_application_configuration_option');
 const ListJDLApplicationConfigurationOption = require('./list_jdl_application_configuration_option');
-const { getTypeForOption, OptionTypes, shouldTheValueBeQuoted } = require('./jhipster/application_options');
+const {
+  getTypeForOption,
+  doesOptionExist,
+  OptionTypes,
+  shouldTheValueBeQuoted
+} = require('./jhipster/application_options');
+
+const logger = require('../utils/objects/logger');
 
 module.exports = { createApplicationConfigurationFromObject };
 
@@ -30,6 +37,10 @@ function createApplicationConfigurationFromObject(configurationObject = {}) {
   const configuration = new JDLApplicationConfiguration();
   Object.keys(configurationObject).forEach(optionName => {
     const optionValue = configurationObject[optionName];
+    if (!doesOptionExist(optionName, optionValue)) {
+      logger.debug(`Unrecognized application option name and value: ${optionName} and ${optionValue}`);
+      return;
+    }
     configuration.setOption(createJDLConfigurationOption(optionName, optionValue));
   });
   return configuration;
--- a/test/spec/exceptions/application_validator.spec.js
+++ b/test/spec/exceptions/application_validator.spec.js
@@ -18,7 +18,12 @@
  */
 
 /* eslint-disable no-new, no-unused-expressions */
-const { expect } = require('chai');
+const chai = require('chai');
+const sinon = require('sinon');
+const sinonChai = require('sinon-chai');
+
+chai.use(sinonChai);
+const { expect } = chai;
 
 const ApplicationValidator = require('../../../lib/validators/application_validator');
 
@@ -34,6 +39,7 @@ const {
   NEO4J
 } = require('../../../lib/core/jhipster/database_types');
 const JDLApplication = require('../../../lib/core/jdl_application');
+const logger = require('../../../lib/utils/objects/logger');
 
 describe('ApplicationValidator', () => {
   let validator;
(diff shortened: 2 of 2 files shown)


lib/core/jdl_application_configuration_factory.js
```diff
@@ -17,12 +17,18 @@
  * limitations under the License.
  */
 
+const logger = require('../utils/objects/logger');
```
MathieuAA, line 20: I'd rather have this import after the others

lib/core/jdl_application_configuration_factory.js
```diff
@@ -36,6 +42,10 @@ function createApplicationConfigurationFromObject(configurationObject = {}) {
 }
 
 function createJDLConfigurationOption(name, value) {
+  if (!doesOptionExist(name, value)) {
+    logger.debug(`Unrecognised application option name: ${name}`);
+    return null;
+  }
```
MathieuAA, line 48: The issue with that is that the JDLAppConfFactory will try to set a nil option.
If it's nil, you have this snippet
```javascript
if (!option) {
    throw new Error('An option has to be passed to set an option.');
}
```
that will throw an error, for the right reasons.
So I suggest rewriting the `createApplicationConfigurationFromObject` function like this:
```javascript
function createApplicationConfigurationFromObject(configurationObject = {}) {
  const configuration = new JDLApplicationConfiguration();
  Object.keys(configurationObject).forEach(optionName => {
    const optionValue = configurationObject[optionName];
    if (!doesOptionExist(optionName, optionValue)) {
      logger.debug(`Unrecognized application option name and value: ${optionName} and ${optionValue}`);
      return;
    }
    configuration.setOption(createJDLConfigurationOption(optionName, optionValue));
  });
  return configuration;
}
```

MathieuAA: Plus, there must be tests for that

SudharakaP: @MathieuAA : If I may ask, that first snippet; 

https://github.com/jhipster/jhipster-core/blob/ea058ca4e1c8eadb067571e5d98cb4927b1ce5d7/lib/core/jdl_application.js#L32-L34

When does it gets called? Because in my tests (I tested running export-jdl command); I didn't see that being called. :smile:

lib/core/jdl_application_configuration_factory.js
SudharakaP: Thanks. Fixed that. :smile:

lib/core/jdl_application_configuration_factory.js
SudharakaP: Ah, I think I've mistakenly removed the test on `jdl_application_configuration.spec.js` which I've added back. I've made the changes as you've suggested. Let me know if you see any further issues. :smile:

MathieuAA: Ah of course, I got confused. If you remove the throw statement, then it's okay.
You understood ;)

test/spec/core/jdl_application_configuration.spec.js
```diff
@@ -100,17 +100,6 @@ describe('JDLApplicationConfiguration', () => {
...
-
-      before(() => {
-        configuration = new JDLApplicationConfiguration();
-      });
-
-      it('should fail', () => {
-        expect(() => configuration.setOption()).to.throw(/^An option has to be passed to set an option\.$/);
-      });
-    });
     context('when setting a new option', () => {
```
MathieuAA, line 103: You've missed a branch here.
Here are the cases (compare those with the `context` paths in the test):
  - The option is passed
    - It didn't exist => it's added
    - It did exit => the value is updated
  - the option isn't passed => **nothing should happen**

The last case is missing (yep, I'm very touchy about hitting the right coverage)

test/spec/exceptions/application_validator.spec.js
```diff
@@ -318,13 +318,6 @@ describe('ApplicationValidator', () => {
           });
         });
       });
-      context('with unknown options', () => {
-        it('should fail', () => {
-          expect(() =>
-            validator.validate(new JDLApplication({ config: { ...basicValidApplicationConfig, toto: 42 } }))
-          ).to.throw(/^Unrecognised application option name: toto\.$/);
-        });
-      });
```
MathieuAA, line 327: I would have kept it but replaced the failure part with "it should send a debug message" (or something like that)

MathieuAA:
Almost done, after the two points it's good to go!

test/spec/core/jdl_application_configuration.spec.js
SudharakaP: Ah, I think I know; I have added this test back. I don't think we have to remove this at all. :smile:

test/spec/exceptions/application_validator.spec.js
SudharakaP: I have replaced it with a different "should not throw an error"; although your message "it should send a debug message" seems maybe more appropriate but I couldn't think of a way to check this debug log message is actually printed in the test suite. Is there a way? :smile:

test/spec/exceptions/application_validator.spec.js
```diff
@@ -319,10 +319,10 @@ describe('ApplicationValidator', () => {
         });
       });
       context('with unknown options', () => {
-        it('should fail', () => {
+        it('should not throw an error', () => {
```
MathieuAA, line 322: Can you check here that the logger's debug function is called?

SudharakaP: @MathieuAA : Got it; done. Sorry I wasn't sure how to do this previously; my knowledge on Chai is limited; but now I learnt a lot :smile:

SudharakaP: And let me know if you see any further issues.

MathieuAA: Ah! I should have mentioned where a similar example could be found. Sorry!!

MathieuAA:
Thank you!

SudharakaP:
You are welcome and thanks for the review. :smile:

mshima:
This PR disables the field validation for everything when the spec is not defined.
This is not a big deal because lexer fails on import with an unknown field, but with blueprint support (unknown field for lexer). This should be fixed.

https://github.com/jhipster/jhipster-core/pull/449 adds an option `ignoreUnknownOptions` that should be added to generator-jhipster.

PR #450, opened 2020-04-05
Forward skipEntityFilesGeneration to entity exporter.

mshima:
  - [ ] Checks are green


mshima:
@MathieuAA let me know if I should keep working on this.

mshima:
Kind of, those options at import isn’t that useful.
But since the options exists...

mshima:
The minor bug is that .jhipster is created when it shouldn't.

MathieuAA:
Yeah I got that. It can be misleading too. Thanks :)

PR #453, opened 2020-04-08
Add support to export unknown configurations.

mshima:
  - [ ] Checks are green

### DIFF
--- a/lib/core/jdl_application_configuration.js
+++ b/lib/core/jdl_application_configuration.js
@@ -17,7 +17,9 @@
  * limitations under the License.
  */
 
-const { OptionNames } = require('./jhipster/application_options');
+const { OptionNames, doesOptionExist } = require('./jhipster/application_options');
+
+const logger = require('../utils/objects/logger');
 
 module.exports = class JDLApplicationConfiguration {
   constructor(config = {}) {
@@ -57,21 +59,21 @@ module.exports = class JDLApplicationConfiguration {
     });
   }
 
-  toString(indent = 0) {
+  toString(indent = 0, exportOptions) {
     const spaceBeforeConfigKeyword = ' '.repeat(indent);
     if (Object.keys(this.options).length === 0) {
       return `${spaceBeforeConfigKeyword}config {}`;
     }
     const spaceBeforeOption = ' '.repeat(2 * indent);
-    const config = getFormattedConfigOptionsString(this.options, spaceBeforeOption);
+    const config = getFormattedConfigOptionsString(this.options, spaceBeforeOption, exportOptions);
     return `${spaceBeforeConfigKeyword}config {
 ${config}
 ${spaceBeforeConfigKeyword}}`;
   }
 };
 
-function getFormattedConfigOptionsString(options, indent) {
-  const filteredOptions = filterOutUnwantedOptions(options);
+function getFormattedConfigOptionsString(options, indent, exportOptions) {
+  const filteredOptions = filterOutUnwantedOptions(options, exportOptions);
   return Object.keys(filteredOptions)
     .sort()
     .map(optionName => {
@@ -81,8 +83,15 @@ function getFormattedConfigOptionsString(options, indent) {
     .join('\n');
 }
 
-function filterOutUnwantedOptions(options) {
-  return filterOutOptionsThatShouldNotBeExported(filterOutOptionsWithoutValues(options));
+function filterOutUnwantedOptions(options, exportOptions = {}) {
+  const { exportUnregisteredConfigs } = exportOptions;
+  options = { ...options };
+  options = filterOutOptionsWithoutValues(options);
+  options = filterOutOptionsThatShouldNotBeExported(options);
+  if (!exportUnregisteredConfigs) {
+    options = filterOutGenericOptions(options);
+  }
+  return options;
 }
 
 function filterOutOptionsWithoutValues(options) {
@@ -104,4 +113,15 @@ function filterOutOptionsThatShouldNotBeExported(options) {
   delete filteredOptions[OptionNames.BLUEPRINTS];
   delete filteredOptions[OptionNames.PACKAGE_FOLDER];
   return filteredOptions;
+}
+
+function filterOutGenericOptions(options) {
+  Object.values(options).forEach(option => {
+    const shouldBeExported = doesOptionExist(option.getName(), option.getValue());
+    if (!shouldBeExported) {
+      logger.debug(`Unrecognized application option name and value: ${option.getName()} and ${option.getValue()}`);
+      delete options[option.getName()];
+    }
+  });
+  return options;
 }
--- a/lib/core/jdl_application_configuration_factory.js
+++ b/lib/core/jdl_application_configuration_factory.js
@@ -22,14 +22,8 @@ const StringJDLApplicationConfigurationOption = require('./string_jdl_applicatio
 const IntegerJDLApplicationConfigurationOption = require('./integer_jdl_application_configuration_option');
 const BooleanJDLApplicationConfigurationOption = require('./boolean_jdl_application_configuration_option');
 const ListJDLApplicationConfigurationOption = require('./list_jdl_application_configuration_option');
-const {
-  getTypeForOption,
-  doesOptionExist,
-  OptionTypes,
-  shouldTheValueBeQuoted
-} = require('./jhipster/application_options');
-
-const logger = require('../utils/objects/logger');
+const GenericJDLApplicationConfigurationOption = require('./generic_jdl_application_configuration_option');
+const { getTypeForOption, OptionTypes, shouldTheValueBeQuoted } = require('./jhipster/application_options');
 
 module.exports = { createApplicationConfigurationFromObject };
 
@@ -37,17 +31,18 @@ function createApplicationConfigurationFromObject(configurationObject = {}) {
   const configuration = new JDLApplicationConfiguration();
   Object.keys(configurationObject).forEach(optionName => {
     const optionValue = configurationObject[optionName];
-    if (!doesOptionExist(optionName, optionValue)) {
-      logger.debug(`Unrecognized application option name and value: ${optionName} and ${optionValue}`);
-      return;
-    }
     configuration.setOption(createJDLConfigurationOption(optionName, optionValue));
   });
   return configuration;
 }
 
 function createJDLConfigurationOption(name, value) {
-  const type = getTypeForOption(name);
+  let type;
+  try {
+    type = getTypeForOption(name);
+  } catch (error) {
+    return new GenericJDLApplicationConfigurationOption(name, value);
+  }
   switch (type) {
     case OptionTypes.STRING:
       return new StringJDLApplicationConfigurationOption(name, value, shouldTheValueBeQuoted(name));
@@ -63,6 +58,6 @@ function createJDLConfigurationOption(name, value) {
       // If this is the case then an option's type isn't one of the cases
       // If there's a new option type, then you should handle it in the switch.
       // If there's no new option type, then you may have made a mistake.
-      throw new Error(`Unrecognized option type: ${type}.`);
+      throw new Error(`Unrecognized option type: ${type} for option ${name}.`);
   }
 }
--- a/lib/core/jdl_application_factory.js
+++ b/lib/core/jdl_application_factory.js
@@ -38,23 +38,20 @@ module.exports = {
  */
 function createJDLApplication(config = {}) {
   const baseConfig = getDefaultConfigForNewApplication(config);
+  let appConfig;
   switch (config.applicationType) {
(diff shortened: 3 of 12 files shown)


murdos:
Just nitpicking on the wording: wouldn't `external`, `additional` or `extra`  options be better than `unknown` options?

mshima:
> Just nitpicking on the wording: wouldn't `external`, `additional` or `extra` options be better than `unknown` options?

I'm not sure, this is internal only for now.
External/additional/extra should be import/export support but only validation is tested for now.

This is part of https://github.com/jhipster/jhipster-core/pull/449.
It became big, so I will split into smaller features to be easier to review.

MathieuAA:
So we're talking about configuration options, right?
What if, and it's just a question, validation is only done unless there's at least one blueprint in the JDLApplication?
The code is getting "heavier", that's why I raise the other possibility.

mshima:
> What if, and it's just a question, validation is only done unless there's at least one blueprint in the JDLApplication?

Are you taking about this PR or blueprints support?

MathieuAA:
Everything. I'm wondering if we took the right decision or we could have made everything simpler.

mshima:
@MathieuAA you should take a look now.
I think `generic_jdl_application_configuration_option` can replace `*_jdl_application_configuration_option`.

lib/exporters/jdl_exporter.js
```diff
@@ -27,10 +27,12 @@ module.exports = {
...
+ * @param {Object}  [options] - Export options.
+ * @param {Boolean} [options.exportNotRegisteredConfigs] - Export configs not registered on definitions.
  */
-function exportToJDL(jdl, path = 'app.jdl') {
+function exportToJDL(jdl, path = 'app.jdl', options) {
   if (!jdl) {
     throw new Error('A JDLObject has to be passed to be exported.');
   }
-  fs.writeFileSync(path, jdl.toString());
+  fs.writeFileSync(path, jdl.toString(options));
```
MathieuAA, line 37: I don't like the fact that we pass options directly to the JDLObject from the exporter, that means there's a coupling between these two (almost independent) parts of JCore.
As I understand the code right now, I think that the toString method takes options with a boolean, might as well have two functions.

lib/core/jdl_application_factory.js
```diff
@@ -38,23 +38,20 @@ module.exports = {
...
+      appConfig = getConfigForUAAApplication(baseConfig);
+      break;
     case ApplicationTypes.MONOLITH:
     default:
-      return new JDLApplication({
-        config: getConfigForMonolithApplication(baseConfig)
-      });
+      appConfig = getConfigForMonolithApplication(baseConfig);
   }
+  return new JDLApplication({ config: appConfig });
```
MathieuAA, line 56: Might as well directly return, unless this is actually a part of the other PR

lib/core/jdl_application_configuration_option.js
```diff
@@ -28,11 +28,19 @@ module.exports = class JDLApplicationConfigurationOption {
...
     return this.value;
   }
 
   toString() {
-    return `${this.name} ${this.value}`;
+    return this._toString(this.value);
+  }
+
+  _toString(value) {
+    return `${this.name} ${value}`;
```
MathieuAA, line 44: why have two methods?
if the second one is to be overridden, might as well have one method and let subclasses override the one

lib/core/jdl_application_configuration_factory.js
```diff
@@ -22,32 +22,27 @@ const StringJDLApplicationConfigurationOption = require('./string_jdl_applicatio
...
   return configuration;
 }
 
 function createJDLConfigurationOption(name, value) {
-  const type = getTypeForOption(name);
+  let type;
+  try {
+    type = getTypeForOption(name);
+  } catch (error) {
+    return new GenericJDLApplicationConfigurationOption(name, value);
```
MathieuAA, line 44: So here generic means unknown? As I understand it, `GenericJDLApplicationConfigurationOption` is a wrapper around options that have unknown types. By the way, what unknown type could there be?

lib/core/jdl_application_configuration.js
```diff
@@ -100,8 +109,19 @@ function filterOutOptionsWithoutValues(options) {
 }
 
 function filterOutOptionsThatShouldNotBeExported(options) {
-  const filteredOptions = { ...options };
+  const filteredOptions = options;
```
MathieuAA, line 112: The reason behind the previous code was to actually duplicate the passed options and not modify it directly to avoid side-effects.

lib/core/jdl_application_configuration.js
```diff
@@ -81,12 +83,19 @@ function getFormattedConfigOptionsString(options, indent) {
     .join('\n');
 }
 
-function filterOutUnwantedOptions(options) {
-  return filterOutOptionsThatShouldNotBeExported(filterOutOptionsWithoutValues(options));
+function filterOutUnwantedOptions(options, exportOptions = {}) {
+  const { exportNotRegisteredConfigs } = exportOptions;
```
MathieuAA, line 87: `exportNotRegisteredConfigs` -> `exportUnregisteredConfigs`

lib/core/jdl_application_configuration.js
```diff
@@ -81,12 +83,19 @@ function getFormattedConfigOptionsString(options, indent) {
...
+  options = { ...options };
+  options = filterOutOptionsWithoutValues(options);
+  options = filterOutOptionsThatShouldNotBeExported(options);
+  if (!exportNotRegisteredConfigs) {
```
MathieuAA, line 91: `exportNotRegisteredConfigs` -> `exportUnregisteredConfigs`

lib/core/jdl_application_configuration.js
```diff
@@ -81,12 +83,19 @@ function getFormattedConfigOptionsString(options, indent) {
...
+    options = filterOutGenericOptions(options, exportOptions);
```
MathieuAA, line 92: The function takes one argument instead of two.

lib/core/jdl_application_configuration.js
mshima: Yes, it's a local function and already is done at the parent function filterOutUnwantedOptions.

lib/core/jdl_application_factory.js
mshima: When I changed to post validation, it's not necessary for now.
Kept because can be needed to pass options to lexer.

mshima: I am revert if you prefer.

test/spec/core/generic_jdl_application_configuration_option.js
```diff
@@ -0,0 +1,66 @@
...
+        jdlConfigurationOption = new GenericJDLApplicationConfigurationOption('foo', 'ba$');
+      });
+      it('should add quotes', () => {
+        expect(jdlConfigurationOption.toString()).to.be.equal('foo "ba$"');
+      });
+    });
+
+    context('when passing value with .', () => {
+      let jdlConfigurationOption;
+      before(() => {
```
mshima, line 48:
```suggestion

      before(() => {
```

test/spec/core/generic_jdl_application_configuration_option.js
```diff
@@ -0,0 +1,66 @@
...
+        jdlConfigurationOption = new GenericJDLApplicationConfigurationOption('foo', '.bar');
+      });
+      it('should add quotes', () => {
+        expect(jdlConfigurationOption.toString()).to.be.equal('foo ".bar"');
+      });
+    });
+
+    context('when passing array', () => {
+      let jdlConfigurationOption;
+      before(() => {
```
mshima, line 58:
```suggestion

      before(() => {
```

test/spec/core/jdl_application_configuration_factory.spec.js
```diff
@@ -111,6 +113,16 @@ describe('JDLApplicationConfigurationFactory', () => {
           expect(createdConfiguration).to.deep.equal(expectedConfiguration);
         });
       });
+      context('containing an unknown configuration', () => {
+        let config;
+        before(() => {
```
mshima, line 118:
```suggestion

        before(() => {
```

test/spec/core/jdl_application_configuration_factory.spec.js
```diff
@@ -111,6 +113,16 @@ describe('JDLApplicationConfigurationFactory', () => {
...
+          config = createApplicationConfigurationFromObject({ toto: 42 });
+        });
+        it('it should add the generic field', () => {
```
mshima, line 121:
```suggestion
        it('should add the generic field', () => {
```

test/spec/exceptions/application_validator.spec.js
```diff
@@ -328,22 +322,6 @@ describe('ApplicationValidator', () => {
...
-        after(() => {
-          loggerDebug.restore();
-        });
-        it('it should send a debug message', () => {
-          expect(loggerDebug).to.have.been.calledOnce;
-          expect(loggerDebug.getCall(0).args[0]).to.equal(
-            'Unrecognized application option name and value: toto and 42'
-          );
-        });
-      });
```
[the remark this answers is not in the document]
mshima: Moved to https://github.com/jhipster/jhipster-core/blob/a27cc6afdd106fd0d0e221c3bf6853332b1380b0/test/spec/exporters/jdl_exporter.spec.js#L138-L143

test/spec/exporters/jdl_exporter.spec.js
```diff
@@ -94,5 +102,108 @@ describe('JDLExporter', () => {
         });
       });
     });
+
+    context('when passing unknown configs', () => {
+      context('without exportNotRegisteredConfigs', () => {
```
mshima, line 107:
```suggestion
      context('without exportUnregisteredConfigs', () => {
```

test/spec/exporters/jdl_exporter.spec.js
```diff
@@ -94,5 +102,108 @@ describe('JDLExporter', () => {
...
+        const PATH = 'app.jdl';
+        let fileExistence;
+        let jdlContent = '';
```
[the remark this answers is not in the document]
mshima:
```suggestion
        let jdlContent;
```

test/spec/exporters/jdl_exporter.spec.js
```diff
@@ -94,5 +102,108 @@ describe('JDLExporter', () => {
...
+        let loggerDebug;
+        before(() => {
```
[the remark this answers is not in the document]
mshima:
```suggestion

        before(() => {
```

test/spec/exporters/jdl_exporter.spec.js
```diff
@@ -94,5 +102,108 @@ describe('JDLExporter', () => {
...
+          expect(loggerDebug.getCall(0).args[0]).to.equal(
+            'Unrecognized application option name and value: toto and foo'
+          );
+        });
+      });
+
+      context('with exportNotRegisteredConfigs', () => {
+        const PATH = 'app.jdl';
+        let fileExistence;
+        let jdlContent = '';
```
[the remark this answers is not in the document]
mshima:
```suggestion
        let jdlContent;
```

lib/exporters/jdl_exporter.js
```diff
@@ -27,10 +27,12 @@ module.exports = {
  * Writes down the given JDL to a file.
  * @param jdl the JDL to write.
  * @param path the path where the file will be written.
+ * @param {Object}  [options] - Export options.
+ * @param {Boolean} [options.exportNotRegisteredConfigs] - Export configs not registered on definitions.
```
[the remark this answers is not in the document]
mshima:
```suggestion
 * @param {Boolean} [options. exportUnregisteredConfigs] - Export configs not registered in definitions.
```

lib/core/jdl_application_configuration.js
mshima:
```suggestion
    options = filterOutGenericOptions(options);
```

lib/core/jdl_application_configuration.js
mshima:
```suggestion
  if (! exportUnregisteredConfigs) {
```

lib/core/jdl_application_configuration.js
mshima:
```suggestion
  const { exportUnregisteredConfigs } = exportOptions;
```

lib/core/jdl_application_configuration_factory.js
mshima: The only different behaviour is list(array/set).
Others (boolean, integer, string) can have the same behaviour.
String have a quoted predefined option. Boolean/Integer are dump.
For JCore 8 I suggest to remove every one and keep only JDLApplicationConfigurationOption with Generic behaviour.

PR #457, opened 2020-04-09
Added support for the blueprints attribute in applications

MathieuAA:
  - [ ] Checks are green

### DIFF
--- a/lib/core/jhipster/application_options.js
+++ b/lib/core/jhipster/application_options.js
@@ -108,7 +108,6 @@ const optionValues = {
   [optionNames.BASE_NAME]: 'jhipster',
   [optionNames.BLUEPRINT]: undefined,
   [optionNames.BLUEPRINTS]: [],
-  [optionNames.BLUEPRINTS]: [],
   [optionNames.BUILD_TOOL]: {
     maven: 'maven',
     gradle: 'gradle'
--- a/lib/core/jhipster/application_options.js
+++ b/lib/core/jhipster/application_options.js
@@ -108,6 +108,7 @@ const optionValues = {
   [optionNames.BASE_NAME]: 'jhipster',
   [optionNames.BLUEPRINT]: undefined,
   [optionNames.BLUEPRINTS]: [],
+  [optionNames.BLUEPRINTS]: [],
   [optionNames.BUILD_TOOL]: {
     maven: 'maven',
     gradle: 'gradle'
--- a/lib/dsl/validator.js
+++ b/lib/dsl/validator.js
@@ -45,6 +45,7 @@ const PATH_PATTERN = /^"([^\/]+).*"$/;
 const REPONAME_PATTERN = /^"((?:http(s)?:\/\/)?[\w.-]+(?:\.[\w.-]+)+[\w\-._~:\/?#[\]@!$&'()*+,;=]+|[a-zA-Z0-9]+)"$/;
 const JWT_SECRET_KEY_PATTERN = /^\S+$/;
 const NUMERIC = /^\d$/;
+const BASIC_NPM_PACKAGE_NAME_PATTERN = /^(@[a-z0-9-][a-z0-9-._]*\/)?[a-z0-9-][a-z0-9-._]*$/;
 
 const configPropsValidations = {
   APPLICATION_TYPE: {
@@ -64,12 +65,12 @@ const configPropsValidations = {
   },
   BLUEPRINT: {
     type: 'NAME',
-    pattern: ALPHANUMERIC_DASH,
+    pattern: BASIC_NPM_PACKAGE_NAME_PATTERN,
     msg: 'blueprint property'
   },
   BLUEPRINTS: {
     type: 'list',
-    pattern: ALPHANUMERIC_DASH,
+    pattern: BASIC_NPM_PACKAGE_NAME_PATTERN,
     msg: 'blueprints property'
   },
   BUILD_TOOL: {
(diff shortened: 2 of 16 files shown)


lib/validators/application_validator.js
```diff
@@ -108,7 +108,9 @@ function checkForValidValues(jdlApplication) {
     'dtoSuffix',
     'entitySuffix',
     'otherModules',
-    'creationTimestamp'
+    'creationTimestamp',
+    'blueprint',
```
mshima, line 112: Blueprint option is deprecated, I think it shouldn’t be added.

lib/core/jhipster/application_options.js
```diff
@@ -108,6 +108,7 @@ const optionValues = {
   [optionNames.BASE_NAME]: 'jhipster',
   [optionNames.BLUEPRINT]: undefined,
   [optionNames.BLUEPRINTS]: [],
+  [optionNames.BLUEPRINTS]: [],
```
mshima, line 111: Duplicated

lib/dsl/validator.js
```diff
@@ -67,6 +67,11 @@ const configPropsValidations = {
     pattern: ALPHANUMERIC_DASH,
     msg: 'blueprint property'
   },
+  BLUEPRINTS: {
+    type: 'list',
+    pattern: ALPHANUMERIC_DASH,
```
mshima, line 72: Can be lower too.

lib/validators/application_validator.js
MathieuAA: Ah! I didn't know that. It will be removed in another commit

lib/dsl/validator.js
MathieuAA: Here's the regex: `/^[A-Za-z][A-Za-z0-9-]*$/`, it takes the lowercase letters into account

lib/validators/application_validator.js
MathieuAA: Wait, deprecated or already not supported?

mshima: Is marked for remove on JHipster 7.
I think we shouldn’t add at all.
blueprint is a single string, blueprints is array.
It doesn’t worth.

lib/dsl/validator.js
mshima: Ops I mean 
```suggestion
    pattern: ALPHANUMERIC_DASH_LOWER,
```

MathieuAA: Ah! you're talking about renaming it! Ah yes, it's a good idea.

lib/validators/application_validator.js
MathieuAA: Alright then!

lib/dsl/validator.js
mshima: ALPHANUMERIC_DASH_LOWER don’t exists now.

lib/validators/application_validator.js
MathieuAA: I'm tagging it "to be removed" for JCore 8

lib/dsl/validator.js
MathieuAA: I thought you meant that we should rename the variable as it's misleading and doesn't tell it also supports the lowercase letters when it actually does.

MathieuAA: alpha numeric could also mean "the 'common' alphabet including uppercase and lowercase letters + numbers" though

mshima: You can use this one:
`((@[a-z0-9-~][a-z0-9-._~]*)\/)?([a-z0-9-~][a-z0-9-._~]*)`
It can be a valid npm package name

MathieuAA: NPM package names are a mess.
https://docs.npmjs.com/files/package.json#name states some rules, but not all of them... There's also [this package](https://www.npmjs.com/package/validate-npm-package-name) from NPM (that seems to be abandoned).

MathieuAA: It seems the package is the officially used one at NPM, so its rules are still used... I'm making a regex out of it for blueprints

mshima: You can take a look at https://github.com/dword-design/package-name-regex

MathieuAA: Yeah but it's wrong. `/^(@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*$/.test('aaa~')` returns true. Do an NPM init with the same test string and see it fail.

MathieuAA: I've opened an issue in the repo BTW.

MathieuAA:
@mshima can you have a look at it and see if I've forgotten something? thanks!

test/spec/jdl/jdl_importer.spec.js
```diff
@@ -1865,5 +1865,60 @@ paginate * with infinite-scroll
...
+
+      it('should return the blueprints attributes in the application', () => {
+        expect(importState.exportedApplications).to.deep.equal([
+          {
+            entities: [],
+            'generator-jhipster': {
+              applicationType: 'monolith',
+              authenticationType: 'jwt',
+              baseName: 'appWithBlueprints',
+              blueprints: ['generator-jhipster-vuejs', 'generator-jhipster-dotnetcore'],
```
mshima, line 1889: I think the generator doesn’t support simple arrays.
Should be `{name: blueprintName}`, version is not required.
https://github.com/jhipster/generator-jhipster/blob/10fa9ba215ede70096d880f804b1e4a0f1fd438b/test/blueprint/scoped-blueprint.spec.js#L63

